PR #5, opened 2015-05-17
added support for tags in put

longnguyen11288:


assets/out
```diff
@@ -35,7 +35,20 @@ if [ -n "${username}" ] && [ -n "${password}" ] && [ -n "${email}" ]; then
 fi
 
 repository=$(jq -r '.source.repository // ""' < $payload)
-tag=$(jq -r '.source.tag // "latest"' < $payload)
+tag=$(jq -r '.params.tag // ""' < $payload)
```
vito: I think we should still support .source.tag in addition to .params.tag (params should take priority if both are specified).

This is used for treating them like branches, which used to be our workflow in some cases (i.e. push to 'latest' vs. 'dev').

longnguyen11288: ok accounted for source and params

assets/out
```diff
@@ -35,7 +35,29 @@ if [ -n "${username}" ] && [ -n "${password}" ] && [ -n "${email}" ]; then
...
+tag_source=$(jq -r '.source.tag // ""' < $payload)
+tag_params=$(jq -r '.params.tag // ""' < $payload)
+
+tag_prefix=$(jq -r '.params.tag_prefix // ""' < $payload)
+
+tag=""
+if [ -n "$tag_params" ]; then
+  tag=$tag_params
+elif [ -n "$tag_source" ]; then
+  tag=$tag_source
```
vito: This won't work since `.source.tag` should be a static tag (i.e. `latest`) whereas `.params.tag` would be a path to a file. So the logic should probably be something like (in pseudocode):

```
if tag_params given
  if tag_params file does not exist
    error
  end

  tag_name="${tag_prefix}$(cat tag_params)"
else if source tag given
  tag_name="${tag_source}"
else
  tag_name=latest
end
```

(could be simplified further if .source.tag defaults to latest as it did before)

longnguyen11288: ahh much cleaner, done

vito:
lgtm, thanks!

Issue #4, opened 2015-05-17
docker-image-resource doesnt play well with private registry


Nopik:
I have some private registry (basically `docker pull registry &&  docker run -p 5000:5000 registry`) and I would like for docker-image-resource to play with it.

In the registry I keep some of my base image which I'm using for all commands, tasks etc. It is basically an ubuntu:14.04 with bunch of stuff pre-loaded (like git, curl, etc.). So, whenever I'm adding task with `image: docker://192.168.59.103:5000/base-ubuntu` it works. But, inside job in concourse I would like to build another image out of that base image of mine. Assuming that I have a folder with `Dockerfile` starting with `FROM base-ubuntu`, from shell it is easy to do:
```
docker pull 192.168.59.103:5000/base-ubuntu
docker tag 192.168.59.103:5000/base-ubuntu base-ubuntu
docker build -t my-image .
docker tag my-image 192.168.59.103:5000/my-image
docker push 192.168.59.103:5000/my-image
```
This sequence isn't really possible now with docker-image-resource. Having the folder done in previous step, my original approach was:
```
- put: my-image
  params:
    build: my-folder/foo
```
And that basically have 2 problems:
1. registry is insecure
2. base-image is not found

I tried to add:
```
  - get: base-image
    resource: base-image
```
with appropriate resource, but it didn't helped, again due to insecure registry.

I'm not sure if `get` should be here at all, maybe not. Either way, I'd love to see docker-image-registry supporting my scenarios, as basically now the only alternative is to have custom docker scripts running against custom docker server.

cjcjameson:
This issue has been closed for a while, so I don't mean to re-open it. But it seems there are multiple people having issues with non-Dockerhub registries. See issue #25

kipz:
I'm also seeing this issue, and my docker registry is not insecure. Concourse seems to ignore my registry and append it to dockerhub. 

E.g. myrepository.com/myimage

resource script '/opt/resource/check []' failed: exit status 1

stderr:
failed to fetch manifest: Get https://registry-1.docker.io/v2/ myrepository.com/myimage/manifests/1.0.0: token auth attempt for registry: https://auth.docker.io/token?account=me&scope=repository%3Amyregistry.com%2Fmyimage%3Apull&service=registry.docker.io request failed with status: 429 Too Many Requests

If I specify the port (443), the pull seems to work.

PR #6, opened 2015-05-18
Added support for extra_args during docker daemon start.

Nopik:
This change, together with `FROM my-repository:5000/my-base-image` approach and
```
source:
  extra_args: --insecure-registry 192.168.59.103:5000
```
solves #4 issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ Tracks and builds [Docker](https://docker.io) images.
 
 * `email`: *Optional.* The email to use when authenticating.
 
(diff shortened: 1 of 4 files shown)


vito:
Generally I'm afraid of people's pipelines being coupled to particular Docker versions (in case they change flags around again), but maybe that's not so bad.

Could we call it `server_args` instead?

Nopik:
Sure. I also added line in readme.

vito:
lgtm

PR #7, opened 2015-06-17
Accept repositories from a private registries

BugRoger:
This will allow to specify repositories in private registries:

```
  - name: rabbit.docker
    type: docker-image
    source:
      repository: docker.mo.sap.corp/concourse/rabbit
```

If the registry host is omitted it will fallback to the default docker.io registy.

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -25,6 +25,13 @@ if [ -z "$repository" ]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


vito:
thanks!

vito:
I had to revert this as it uses syntax not available in the image, which its scripts via `sh`, not `bash`. Sorry!

PR #8, opened 2015-07-06
Add support for `--no-cache` option for `docker build`.

oppegard:


vito:
What does passing --no-cache do? Seems like it'd not have any effect anyway since the container is fresh every time.

oppegard:
Correct, passing --no-cache won't have any effect since the container is fresh. However, I figured we might as well pass the option down to docker in the event things change in the future (volume support?),

vito:
Gotcha. That probably won't be changing; when we do the volume stuff we ideally won't be changing semantics like that (otherwise it's kind of a leaky optimization).

oppegard:
Want to merge as-is, or should I re-submit without the --no-cache option?

vito:
Removing the `--no-cache` would be great; it's a bit misleading as-is. Maybe just have a `cache` option which defaults to true?

oppegard:
Renamed parameter to  `cache` and squashed to a single commit.

vito:
thanks!

PR #9, opened 2015-07-06
Support `docker import` for pushing exported rootfs

mhoran:
Buildpacks generates a rootfs.tar.gz for CF rootfses. We then need to push this tarball to Dockerhub.

The `docker load` command does not support loading tarballs that were exported via `docker export`. It only supports those saved via `docker save`. This PR adds support for importing tarballs generated by `docker export` to be pushed to Dockerhub.


vito:
thanks!

PR #10, opened 2015-08-01
Support private registry

d:
This PR adds a `source.registry` configuration to set a private v1 Docker registry to push to / pull from.

We've been using this in our pipeline for a while.

Thoughts?

### DIFF
--- a/assets/common.bash
+++ b/assets/common.bash
@@ -13,7 +13,15 @@ start_docker() {
       mount -n -t cgroup -o $d cgroup /sys/fs/cgroup/$d
   done
 
-  docker $1 -d >/dev/null 2>&1 &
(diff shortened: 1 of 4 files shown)


assets/common.bash
```diff
@@ -23,6 +32,31 @@ start_docker() {
...
+extract_registry() {
+  local repository="${1}"
+
+  echo "${repository}" | cut -d/ -f1
+}
+
+extract_repository() {
+  local long_repository="${1}"
+
+  echo "${long_repository}" | cut -d/ -f2
```
vito: i think this'll just take `bar` from `foo.com/bar/baz` - you want `bar/baz` no?

vito:
Thanks!

PR #11, opened 2015-08-13
Allow skipping docker pull during `get` step [#100994844]

monkeyherder:


assets/in
```diff
@@ -37,22 +37,40 @@ fi
...
+skip_download="$(jq -r '.params.skip_download // false' < $payload)"
 
 mkdir -p $destination
 
-docker save -o ${destination}/image "$image"
-
-if [ "$rootfs" = "true" ]; then
-  docker run --cidfile=container.cid "$image" echo container created
-  docker export $(cat container.cid) > ${destination}/rootfs.tar
+if ! $skip_download
```
xoebus: This should be of the form:

```
if [ "$skip_download" = "true" ]; then
```

to avoid subtle `bash` behavior.

monkeyherder:
@xoebus Cleaned up to avoid `bash` subtleties.

assets/in
```diff
@@ -37,22 +37,39 @@ fi
...
+  if [ "$rootfs" = "true" ]; then
+    docker run --cidfile=container.cid "$image" echo container created
+    docker export $(cat container.cid) > ${destination}/rootfs.tar
+  fi
+else
+  if private_registry "${repository}"
+  then
+    registry=$(extract_registry $repository)
+    repository=$(extract_repository $repository)
+    json_selector="."
```
vito: is docker really inconsistent within its own "v1" API?

:'(

monkeyherder:
Yeah, docker really is incompatible with its own "v1" API. See https://github.com/docker/docker-registry/issues/517#issuecomment-52786954 . While the discussion is about the `tags` endpoint, rather than `tags/<version>`, we saw the same sort of inconsistent behavior.

PR #13, opened 2015-09-02
Fix example config in README.md to include source repository

sclevine:
This field was left out of the example, even though it is required.


xoebus:
Thanks!

PR #16, opened 2015-10-26
Dynamic protocol

### LINKED ISSUE #15: Protocol port of source config
Would you guys be opposed to being able to select a protocol as part of source config?

My issue right now is that I'm using a private docker repo that uses https (not http)

mehtaphysical:
resolves #15 

* set http/https depending on the availability of the v1/_ping endpoint
* add `-k` to curl_opts if `--insecure-registry` is set

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -40,7 +40,7 @@ else
   registry=registry.hub.docker.com
 fi
 
-if [[ "$server_args" == *"--insecure-registry=bosh.getlocalmotion.com"* ]]; then
+if [[ "$server_args" == *"--insecure-registry="* ]]; then
   curl_opts="${curl_opts} -k"
 fi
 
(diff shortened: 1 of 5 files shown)


vito:
Sorry for the delay on this. Have you manually verified that this works? There was at least one point in time where the scripts used bash-specific features, but it looks like you've fixed it.

mehtaphysical:
no worries.

It seems to be working. I've been using it without issue for a few weeks now. 

Yeah at first i wrote it as bash. I've updated it since then. Sorry about that.

assets/check
```diff
@@ -35,23 +36,29 @@ fi
...
-curl ${curl_opts} "${uri_root}/v1/repositories/${repository}/images" > $all_images
+shortened_image=$(jq -r "try map(select(.name == $(echo $tag | jq -R .)))[].layer" < $tags)
 
-latest_image=$(jq -r "map(select(.id | startswith($(echo $shortened_image | jq -R .))))[0].id // \"\"" < $all_images)
+if [ -n shortend_image ]; then
+  all_images=$(mktemp /tmp/resource-in-tags.XXXXXX)
+  curl ${curl_opts} "${uri_root}/v1/repositories/${repository}/images" > $all_images
+  latest_image=$(jq -r "map(select(.id | startswith($(echo $shortened_image | jq -R .))))[0].id // \"\"" < $all_images)
+else
+  latest_image=$(jq -r "$(echo $tag) | jq -R .")
```
vito: What are these changes for? Looks unrelated to the rest (also this line has broken syntax).

vito:
So I ended up rewriting `check` to support v2 endpoints in 5d6c1ddfdf84e1799dc7866dfb9d7ec5fb2dfd54. Along the way I added an explicit `source.insecure_registries` property, and `check` will try `https` and fall back on `http`. Also `in` no longer does the `curl` so it shouldn't need anything special.

If you want to give it a try, you can point your instance at `docker:///concourse/docker-image-resource`. Note that a few things have changed (see the README). Notably it versions via digests now.

PR #20, opened 2016-01-15
Avoid eating memory via large tmpfs.

Mook-as:
Use a local disk image instead of a tmpfs mount for /var/lib/docker to avoid consuming all of the memory on the worker.

Fixes concourse/concourse#255 [#111720876]
Requires concourse/core-busyboxplus#1

Needed to add a cleanup to avoid leaking the mount, but at least that's not all of the memory even if it leaks.

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -13,8 +13,18 @@ start_docker() {
       mount -n -t cgroup -o $d cgroup /sys/fs/cgroup/$d
   done
 
+  for i in $(seq 0 100) ; do
+    mknod -m 660 /dev/loop${i} b 7 ${i}
+  done
+
   mkdir -p /var/lib/docker
-  mount -t tmpfs -o size=100G none /var/lib/docker
+  image=$(mktemp $PWD/docker.img.XXXXXXXX)
+  dd if=/dev/zero of=${image} bs=1 count=0 seek=100G
+  mkfs.ext4 -F ${image}
+  losetup -f ${image}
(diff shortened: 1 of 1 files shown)


assets/common.sh
```diff
@@ -13,8 +13,18 @@ start_docker() {
...
+  for i in $(seq 0 100) ; do
+    mknod -m 660 /dev/loop${i} b 7 ${i}
+  done
+
   mkdir -p /var/lib/docker
-  mount -t tmpfs -o size=100G none /var/lib/docker
+  image=$(mktemp $PWD/docker.img.XXXXXXXX)
+  dd if=/dev/zero of=${image} bs=1 count=0 seek=100G
+  mkfs.ext4 -F ${image}
+  losetup -f ${image}
```
vito: Could this be changed to `losetup -f --show` instead of grepping for the image? That'll atomically acquire and print out which loopback device was picked. Hopefully that flag's available in the image, not sure. With that change you also no longer need the randomly generated image name.

Mook-as: Unfortunately, no; busybox `losetup` doesn't support `--show`.  I actually tried that one first because I was following the [code in baggageclaim][btrfs.go] you suggested. :|  This (and the random filename) was all to workaround that lack.

[btrfs.go]: https://github.com/concourse/baggageclaim/blob/7fdf04194a2e8fa28e26c03a03756f7f3c389655/fs/btrfs.go#L33-L51

vito: Aw, too bad. `mktemp` works then.

PR #21, opened 2016-01-18
Force set tags when loading from files because they can be the same

Mook-as:
`docker save` will store the tag name that the saved image was from, and `docker load` will restore it.  If it is the same as the desired tag name, `docker tag` will refuse to re-tag the image to the desired tag and fail.

### DIFF
--- a/assets/out
+++ b/assets/out
@@ -84,7 +84,7 @@ rootfs="$(jq -r '.params.rootfs' < $payload)"
 
 if [ -n "$load_base" ]; then
   docker load -i "${load_base}/image"
-  docker tag \
+  docker tag --force \
     "$(cat "${load_base}/image-id")" \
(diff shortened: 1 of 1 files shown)


vito:
Now seeing this output in my builds:

```
Warning: '--force' is deprecated, it will be removed soon. See usage.
```

May want to roll this back.

Mook-as:
Argh.
Looking at their [deprecation notices], it looks like `--force` is now always true as of docker 1.10 and the flag will be removed in 1.12.  So this PR was _only_ useful before 1.10 shipped. Sigh.

Sorry about that.

[deprecation notices]: https://docs.docker.com/engine/deprecated/#f-flag-on-docker-tag

vito:
Hah, no worries. I'll revert it.

PR #22, opened 2016-02-10
enable using a dockerfile

pixeleet:


assets/out
```diff
@@ -94,7 +96,7 @@ if [ -n "$build" ]; then
     docker_pull "${repository}:${tag_name}"
   fi
 
-  docker build -t "${repository}:${tag_name}" "$build"
+  docker build -t "${repository}:${tag_name}" -f "$dockerfile" "$build"
```
vito: Will this work? Seems like it'd result in `docker build -t foo:bar -f Dockerfile ./foo`, when you really want `-f ./foo/Dockerfile` for the current behavior.

`docker build -h` makes it sound like it should be a path to the file, and isn't relative to the path:

```
  -f, --file=                     Name of the Dockerfile (Default is 'PATH/Dockerfile')
```

pixeleet:
Good question. I was wondering about that too, and I think it's relative.
https://docs.docker.com/engine/reference/commandline/build/#specify-dockerfile-f

At least it has to be part of the build context you're uploading, since I'm not sure, I've just updated it to only use a file if one is explicitly set, I think that should be fine.

vito:
Based on those docs I think it's relative to the working directory, not to the build directory (all of those examples build `.` which makes it work either way).

Regardless I think the updated behavior makes sense; it'll also support the `Dockerfile` living outside the build dir, which seems like a good use case to support.

Merging; thanks!

Issue #27, opened 2016-03-15
assets/in script does not handle images built using "scratch" repo


robdaemon:
Given this pipeline configuration, builds always fail because the source image uses `scratch` as a `FROM`, so there is no `/etc/passwd` file.

Pipeline configuration:

```yaml
jobs:
  - name: docker-images
    plan:
      - get: nats-image-in
      - put: nats-image-out

resources:
  - name: nats-image-in
    type: docker-image
    source:
      repository: nats
      tag: 0.7.2

  - name: nats-image-out
    type: docker-image
    source:
      insecure_registries: ["192.168.41.159:5000"]
      repository: 192.168.41.159:5000/my-repo/nats
      tag: mytag
```

Bug is around here: https://github.com/concourse/docker-image-resource/blob/747c43afda713b0447f3327cd312985f568f6d8d/assets/in#L62-L66

`print-metadata` should handle this scenario gracefully, these are perfectly valid Docker images.

vito:
Should be fixed now that we merged your PR. Thanks!

PR #28, opened 2016-03-18
Handle when /etc/passwd is missing because the image is built using s…

robdaemon:
…cratch or another base layer that does not provide that file.

### DIFF
--- a/cmd/print-metadata/main.go
+++ b/cmd/print-metadata/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


robdaemon:
This resolves #27

vito:
Not a fan of assuming failure to run `docker run` is due to missing `/etc/password` - it could be any number of things. Instead https://github.com/concourse/docker-image-resource/tree/master/cmd/print-metadata should probably be made durable to this.

robdaemon:
I was debating where to handle that. I can make print-metadata more resilient instead.

vito:
Cool, thanks for looking into this!

robdaemon:
Okay, updated the PR. Handling the lack of /etc/passwd and parsing of /etc/passwd as a recoverable error instead of a panic condition.

robdaemon:
Is there anything else that needs to be done on this PR so it can be merged upstream? I'm currently running my fork in production, and I'd like to stick to upstream.

PR #31, opened 2016-04-13
Add tag_as_latest

shinji62:
Added a new option for output resource.
New params `tag_as_latest` will tag whatever we already tag as latest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,9 @@ version is the image's digest.
   prepended with this string. This is useful for adding `v` in front of version
   numbers.
 
+* `tag_as_latest`: *Optional.*  Default `false`. If true, the pushed image will be tag as latest too and tag will be push.
+
+
 ## Example
 
 ``` yaml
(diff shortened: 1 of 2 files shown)


shinji62:
@vito, @mariash any thought on this ?

mariash:
Are you using `tag` for something non-latest and want to tag it as `latest` as well? I wonder if it is better to convert `tag` to the list of `tags`?

shinji62:
Humm not sure if is better, but docker do not tag the last pushed image as latest automatically, so it can be painful when you use some tool like concourse when for example in another pipeline we want to use the latest tag.

I think latest tag is kind of special tag, I don't see the meanings to support more than two tag.
Version tag (0.0.0) and latest should be the only supported tag.

Furthermore tag field is expected a file which is good which is convenient when you use with semver and so on.

That's why I implemented is this way, basically I want to push this image as the tag v.0.0.0 and this images is the latest pushed

Issue #33, opened 2016-04-18
Resource tries to pull image with incorrect digest


martinsbalodis:
Hi!
I am running `fly execute` but it fails while pulling a docker image. From the output I can see that it tries to pull a version of a container that has never existed in the repository (private http repository).

Here is the task.yml part:
```yml
...
image_resource:
  type: docker-image
  source:
   repository: 1.2.3.4:5000/my-image
   insecure_registries: ["1.2.3.4:5000"]
   email: martins@example.com
   username: martins
   password: pass
...
```

Fly command:
```sh
fly -t lc execute -c task.yml
```

Output of the fly command:
```
...
initializing
WARNING: login credentials saved in /root/.docker/config.json
Login Succeeded
Pulling 1.2.3.4:5000/my-image@sha256:59795eb87abd170f1b97f82f5838db5637890962d7bd623f5ce04c7833ef29f0...
Error response from daemon: manifest unknown: manifest unknown

Pulling 1.2.3.4:5000/my-image@sha256:59795eb87abd170f1b97f82f5838db5637890962d7bd623f5ce04c7833ef29f0 (attempt 2 of 3)...
Error response from daemon: manifest unknown: manifest unknown

Pulling 1.2.3.4:5000/my-image@sha256:59795eb87abd170f1b97f82f5838db5637890962d7bd623f5ce04c7833ef29f0 (attempt 3 of 3)...
Error response from daemon: manifest unknown: manifest unknown


Failed to pull image 1.2.3.4:5000/my-image@sha256:59795eb87abd170f1b97f82f5838db5637890962d7bd623f5ce04c7833ef29f0.resource script '/opt/resource/in [/tmp/build/get]' failed: exit status 1
errored
```

The docker registry doesn't contain a version of this image that has sha256 like `59795eb87abd`. I also checked the directory where the images are stored to make sure.

vito:
Don't really know what to tell ya. We collect these versions from the registry API and then just go back and try to fetch them from the same registry. Sounds like your registry is broken or inconsistent? Are you running a v2 registry?

databus23:
I've been hit by this too, today when we upgraded from from 2.3.0 to 2.4.0.

@martinsbalodis Are you be any chance running a registry version >= 2.3.1?

I think this change is causing the problem: https://github.com/docker/distribution/commit/d7eb5d118a6a14a6f320062296b1808506c35241

[...]

PR #34, opened 2016-04-19
Explicitly request v2 schema 2 manifests

databus23:
This is a fix for #33
This fixes discovering digests on docker distribution >= 2.3.1 by explicitly requesting schema v2 manifests.
Before distribution 2.3.1 the v2 schema 1 manifests contained the schema 2 digest by accident which was fixed with this commit https://github.com/docker/distribution/commit/d7eb5d118a6a14a6f320062296b1808506c35241.
docker pull (at least on recent versions of docker) only works with v2 digest. So this only worked by accident until it was fixed.

### DIFF
--- a/cmd/check/main.go
+++ b/cmd/check/main.go
@@ -44,11 +44,10 @@ func main() {
 	manifestURL, err := ub.BuildManifestURL(repo, tag)
 	fatalIf("failed to build manifest URL", err)
 
-        manifestRequest, err := http.NewRequest("GET", manifestURL, nil)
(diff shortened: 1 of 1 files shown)


vito:
Will this continue to work on registries < 2.3.1?

databus23:
It should. But I wasn't able to test it.
IIRCC the registry ignores Accept headers it does not understand. So I would assume older version the registry not aware of schema 2 would just continue to return a schema 1 manifest.

vito:
Thanks! Merging this in now that things are stable again (CI was red for a bit).

Issue #40, opened 2016-05-09
Manifest fetch fails with 406 error


masonoise:
I'm getting `failed to fetch digest: 406 Not Acceptable` that I've narrowed down to the `check` command. The current theory is that our registry is returning type `application/json` while check is looking for `application/vnd.docker.distribution.manifest.v2+json`. See this line: https://github.com/concourse/docker-image-resource/blob/master/cmd/check/main.go#L49

If the header can be configurable, or simply add `application/json`, that would be great.

Our private registry is an Artifactory Docker-type repository, which may explain why the type is different than what's expected.

databus23:
Ok, looks like its a regression introduced by #34, which fixed the resource for docker distribution >=2.3.1.

Could you check the Content-Type returned by your registry:
```
curl -I https://$HOST/v2/$ORG/$repo/manifests/$TAG
```

databus23:
Also, what version of Artifactory are you running?
V2 Schema 2 should be supported since 4.4.3 according to https://www.jfrog.com/jira/browse/RTFACT-9056

masonoise:
Yeah, I checked the registry and it is sending back `Content-Type: application/json`.

We are running Artifactory 4.4.2 rev 40093. Sounds like maybe 4.4.3 would help so I'll see when an upgrade might be planned.

masonoise:
I did confirm that changing that line to `manifestRequest.Header.Add("Accept", "application/json")` fixes the problem, so the accept-header is the issue. I could submit a PR adding that header to the existing one, but if you want to look into making the header configurable that's probably a better long-term solution. Let me know -- thanks.

databus23:
While I'm not in charge of this resource I'm not convinced that making this configurable is the way to go. Ideally the resource should implement the same behaviour as the docker daemon itself to fetch the correct manifest. Would be good to know what docker is actually sending to get the correct manifest in all cases. Unfortunately the docker code base is quite big and not easy to understand.

masonoise:
I added `application/json` to the accept header and testing seems good so I submitted a PR for the change. I'm not entirely sure how to do a complete build of everything in this package to create a new Dockerfile but building and testing the check command was straightforward. Thanks.

databus23:
Unable to figure out from code what docker is doing itself I used `mitmproxy` to take a closer look:
This is what I get when doing docker pull alpine (docker 1.11.1):
```
GET /v2/library/alpine/manifests/latest HTTP/1.1
[...]

PR #41, opened 2016-05-10
Added application/json to manifestRequest Accept header

masonoise:
As described in issue #40 a Docker registry in Artifactory 4.4.2 (and probably earlier) responds to the manifest request with content-type `application/json`, which results in a 406 return code. This change adds that content-type to the Accept header. Have tested against our Artifactory registry and also against the public Docker hub and both work.

### DIFF
--- a/cmd/check/main.go
+++ b/cmd/check/main.go
@@ -46,7 +46,8 @@ func main() {
 
 	manifestRequest, err := http.NewRequest("GET", manifestURL, nil)
 	fatalIf("failed to build manifest request", err)
-	manifestRequest.Header.Add("Accept", "application/vnd.docker.distribution.manifest.v2+json,application/json")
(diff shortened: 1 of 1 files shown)


cmd/check/main.go
```diff
@@ -46,7 +46,7 @@ func main() {
 
 	manifestRequest, err := http.NewRequest("GET", manifestURL, nil)
 	fatalIf("failed to build manifest request", err)
-	manifestRequest.Header.Add("Accept", "application/vnd.docker.distribution.manifest.v2+json")
+	manifestRequest.Header.Add("Accept", "application/vnd.docker.distribution.manifest.v2+json,application/json")
```
vito: Should this be an additional `.Add("Accept", ...)` instead?

masonoise: Based on the findings in issue #40 I suspect it should -- that would map the behavior that Docker seems to be doing. I'll update.

vito:
Looks good - thanks!

PR #37, opened 2016-05-04
Amazon ECR Support

eploko:
PoC implementation as outlined in #36

PR #52, opened 2016-06-29
Alternative ECR Support

jcderr:
In light of the concern from @vito on #37, specifically in terms of getting credentials from multiple sources, I reworked it slightly to support a scenario where the instance has an IAM profile and thus the user will not be providing any credentials at all.

This will require vendoring the awscli, but I profess to not understanding the innards of Concourse quite well enough to see how that would be done.

In all other circumstances where a user would be providing credentials, I'd defer to concourse/concourse#291

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -82,6 +82,21 @@ stop_docker() {
   wait $pid
 }
 
+ecr_registry() {
+    local repository="${1}"
+
+    if echo "${repository}" | grep -q -e '\.ecr\.[^\.]*\.amazonaws\.com' ; then
+        return 0
+    fi
+
+    return 1
+}
+
+ecr_registry_region() {
+    local registry="${1}"
+    echo "${registry}" | cut -d. -f4
+}
+
 private_registry() {
   local repository="${1}"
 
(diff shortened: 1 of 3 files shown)


kimeberz:
Thanks for the PR! We'll get this merged ASAP.

Only qualms (this is @vito btw):

- What I meant by "multiple sources of credentials" wasn't more about resources having a single code path for handling the request. But this is something I'm willing to sacrifice for now.
- In light of that, let's re-add the config for AWS, but as explicit `aws_access_key` and `aws_secret_key` parameters rather than overloading `username` and `password` and pattern-matching the repository. I actually don't like depending on the IAM rules for this as that means you've configured your workers in a special way, which goes against Concourse's ethos.
- We should update the Dockerfile to include the `aws` CLI. But we can do that ourselves once the previous point is changed, since it may be a bit of a rabbit hole.

- @vito & @kimeberz

jcderr:
IAM profiles are a somewhat common corporate requirement. Not supporting them will mean that I must abandon Concourse.

vito:
Then that can be also supported; I'm not against it in principle, I just don't want to require people to use IAM profiles, and I don't want to magically overload `username` and `password` based on a regexp on the repository.

vito:
Picked this up myself; should be in the next release.

PR #56, opened 2016-07-16
Another aws ecr support attempt

ness2u:
I didn't want to spend the time figuring out how to get the aws cli baked into the image; that will need to get done before I can test this.


vito:
Looks good. Do we need to do something for `check` too though?

ness2u:
I'm not a go developer; that check code is my first commit with golang... feedback appreciated.

I couldn't get godep to update after adding the aws sdk dependency, that will need to be looked at by someone more experienced.
`go get -u github.com/aws/aws-sdk-go` will add the missing deps.

ness2u:
Got it all working without the aws cli or python.

oppegard:
@ness2u @vito I'm not sure if you were aware, but should we consider using AWS's Docker "credential helper" for ECR? https://github.com/awslabs/amazon-ecr-credential-helper

I assume they'll keep it up to date, fix bugs, add new features, etc. And it uses an official Docker cli interface. In the long run I figure it will be less maintenance and breakage.

vito:
@oppegard That looks pretty nice - we should definitely use it for `/in` and `/out`.

`/check` will need its own deal though. And it would be nice (@ness2u) if it didn't diverge so much from the normal code path - can we just use https://github.com/awslabs/amazon-ecr-credential-helper/blob/master/ecr-login/ecr.go#L44 and log in as normal?

oppegard:
@ness2u Here are Docker's credential helpers with some docs on how they work: https://github.com/docker/docker-credential-helpers

vito:
Decided to just pick this up myself, using Amazon's credential helper, since it took a bit of knowhow and dealing with boring Go vendoring stuff. Should be in the next release. Thanks!

Issue #46, opened 2016-06-08
Caching is broken with docker 1.10


mariash:
By ellerycrane in slack:

Because of the changes in docker 10, the only way to preserve the build cache is to use `docker save` of the image ​*from the place it was built*​, and then use `docker load` on the output of the former when building it again elsewhere.
the `docker-image-resource` does the `docker build` step in the `/out` script,
and has an option to use `save=true` in the `/in` script
However, in the context of `/in`, it will do a fresh `docker pull` which will not contain the build cache we want to preserve. So instead, I'd like to take the image built as part of the `/out` script and use it to update the latest version of the resource, rather than doing `/in`

See more: https://github.com/docker/docker/issues/20316

ellerycrane:
To summarize the issue: 

With Docker 1.10, they made it so that pulling an image no longer pulls down the build cache for that image. This means that the only way to make use of Docker image layer caching, for both improved build times and reduced i/o time when pulling/pushing, is to make all subsequent builds of an image happen on the same system that did previous builds.

This is a problem for Concourse because no preservation of state happens between builds, so the build cache is lost. Pulling the image no longer retrieves the build cache, so every build must completely rebuild every layer in the docker image and push them all to the remote registry. (Incidentally, this also means that the `cache:true` parameter no longer works; it will pull down the latest image, but it won't contribute anything to the build except additional download time).

[This blog post](http://blog.runnable.com/post/145362675491/distributing-docker-cache-across-hosts) describes the underlying Docker issues in detail, and offers a potential work around: In Docker 1.11, they made it so that you can export the build cache by using `docker save <IMAGE> $(docker history -q <IMAGE>) > image.tar` and then loading that image before a build by using `docker load < image.tar`. The existing `docker-image-resource` code already largely supports this type of mechanism, and could be tweaked to work around the problem.

The biggest challenge I see is how to take the outputs from `/out`, where the image is built and `docker save` could be used to preserve the build cache, and funnel them into the same location as the outputs from `/in`.  

[...]

PR #59, opened 2016-07-27
Cache tag parameter

kichik:
This pull request allows specifying a different tag to pull for caching. This way caching can be used even when a new tag is built. For our case we use a new image tag for each git commit. As it is right now, we can't use `cache` in this case. With this pull request we'd be able to use `tag_as_latest: True` and `cache_tag: latest` to enable caching and cut down build times.

### DIFF
--- a/assets/out
+++ b/assets/out
@@ -73,6 +73,7 @@ load=$(jq -r '.params.load // ""' < $payload)
 load_base=$(jq -r '.params.load_base // ""' < $payload)
 build=$(jq -r '.params.build // ""' < $payload)
 cache=$(jq -r '.params.cache' < $payload)
+cache_tag=$(jq -r '.params.cache_tag // "${tag_name}"' < $payload)
 dockerfile=$(jq -r '.params.dockerfile // ""' < $payload)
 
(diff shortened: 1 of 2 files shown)


veered:
This would be extremely helpful!

vito:
Looks good, thanks!

assets/out
```diff
@@ -73,6 +73,7 @@ load=$(jq -r '.params.load // ""' < $payload)
 load_base=$(jq -r '.params.load_base // ""' < $payload)
 build=$(jq -r '.params.build // ""' < $payload)
 cache=$(jq -r '.params.cache' < $payload)
+cache_tag=$(jq -r '.params.cache_tag // "${tag_name}"' < $payload)
```
databus23: This does not work as intended. ${tag_name} is not expanded inside single quotes.

databus23:
While this change makes sense in general keep in mind that the build caching is broken since docker 1.10. (See https://github.com/concourse/docker-image-resource/issues/46 for details).

kichik:
I can confirm this works with Concourse 2.0.1 and pulls the right image but doesn't yet work as intended due to #46 as @databus23 said.

Thanks!

PR #62, opened 2016-08-22
Fix build cache

databus23:
This PR fixes the broken build cache reported in #46.

It uses https://github.com/tonistiigi/buildcache to export the build cache after building a docker images and re-applies it on subsequent builds when `cache` is set to `true`.
The exported build cache metadata is very small (<10KB) and I chose to use the image's repository to store the build cache in a tiny, separate docker images as it requires no additional configuration on the resource. I realise that this is not ideal but I don't know where else to store the metadata to share it between successive `put` operation without complicating things massively.


vito:
Looks great, thanks!

ashb:
This looks very useful but I'm not quite sure how to use this feature.

I tried just setting `param: {cache: true}` but this then tried to pull the image tag it's about to build -- which doesn't exist.

Is there some easy way to say "refer to the tag output in the previous build"? And how do you bootstrap this the very first time? If the  "${repository}:${cache_tag}-buildcache" doesn't exist (which it won't first time) then I can't run with cache:true, but if `cache:true` isn't set then how is the -buildcache tag ever built and pushed?

PR #63, opened 2016-08-24
support docker build-arg

### LINKED ISSUE #55: put should support build-arg
Docker 1.9 introduced [build-arg](https://docs.docker.com/engine/reference/builder/#/arg), it could be nice to support this feature. 

Something like this (similar to ca-certs)
build-arg: Optional. An array of objects with the following format:
```yaml
build-arg:
- key: maintainer
  value: vito@concourse.org
- key: ruby-gem-mirror
  value: https://my-mirror.com/
```

o-orand:
2 ways to set:
 * using param in put
 * using a yaml file

When both set, the result contains all build arguments (may have duplicate!).

close #55

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,21 @@ version is the image's digest.
 
 * `tag_as_latest`: *Optional.*  Default `false`. If true, the pushed image will be tag as latest too and tag will be push.
 
+* `build_args`: *Optional.*  Default ``. List of docker build arguments. Format: `<key>: <value>`.
+
+    ```yaml
+      build_args:
+        argInt: "1"
+        argBool: "true"
+        argEmail: me@yopmail.com
+    ```            
+* `build_args_file`: *Optional.*  Default ``. A Yaml filename containing docker build arguments.  
+File content sample:
+
+    ```yaml
+    { "argEmail": "me@yopmail.com", "argInt": "1", "argNewBool": "false" }
+    ```            
+
 
 ## Example
 
--- a/assets/out
+++ b/assets/out
@@ -50,6 +50,8 @@ tag_source=$(jq -r '.source.tag // "latest"' < $payload)
 tag_params=$(jq -r '.params.tag // ""' < $payload)
 tag_prefix=$(jq -r '.params.tag_prefix // ""' < $payload)
 need_tag_as_latest=$(jq -r '.params.tag_as_latest // "false"' < $payload)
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -143,6 +143,21 @@ version is the image's digest.
...
 
+* `build_args`: *Optional.*  Default ``. List of docker build arguments. Format: `<key>: <value>`.
+
+    ```yaml
+      build_args:
+        argInt: "1"
+        argBool: "true"
+        argEmail: me@yopmail.com
+    ```            
+* `build_args_file`: *Optional.*  Default ``. A Yaml filename containing docker build arguments.
```
vito: JSON, not YAML, right?

assets/out
```diff
@@ -100,10 +102,19 @@ elif [ -n "$build" ]; then
     docker_pull "${repository}:${cache_tag}"
   fi
 
+  expanded_build_args=""
+  if [ -n "$build_args" ]; then
+    expanded_build_args=$(echo $build_args|jq -r 'with_entries(.key |= "--build-arg " + . )|with_entries(.key = .key + "=" +.value)|keys|join(" ")')
```
vito: I'm a bit worried about escaping semantics here (and in the call site), e.g. if the value has characters like `!` or spaces. Have you tried that out?

o-orand: You're right, this doesn't support `space` neither `!` :disappointed: . I'm going to see how to escape special characters. I'm not sure it's possible using jq.

README.md
o-orand: right, file is parsed with jq...

assets/out
vito: To be honest it's probably time to rewrite `out` in Go; it's been getting pretty complicated over time. I'm not sure if you wanna bite that bullet as part of this PR though. We've at least done that for `check` already.

o-orand: Indeed, it could be nice to rewrite it in Go, but I'm not ready to do it yet. Maybe later :smile:

vito:
I'll pull this in as-is even though we know it'll break with fancier arguments. Thanks!

I'll update the README to say the file must be JSON, not YAML.

Paging Dr. @zachgersh for the Go rewrite. 😉

zachgersh:
@vito page received but you picked git first 👯

vito:
did not - docker image resource is way more important. i said git over bosh deployment resource :)

PR #66, opened 2016-09-04
Allow use of ECR urls in FROM

donaldguy:
Conditionally adds url from Dockerfile to auths per
awslabs/amazon-ecr-credential-helper#9

A `fly execute -c ci/build.yml --output=out/ && docker build out` version
is pushed to https://hub.docker.com/r/donaldguy/concourse-docker-image-resource
and I tested it on my server

### DIFF
--- a/assets/out
+++ b/assets/out
@@ -76,7 +76,7 @@ load_base=$(jq -r '.params.load_base // ""' < $payload)
 build=$(jq -r '.params.build // ""' < $payload)
 cache=$(jq -r '.params.cache' < $payload)
 cache_tag=$(jq -r ".params.cache_tag // \"${tag_name}\"" < $payload)
-dockerfile=$(jq -r '.params.dockerfile // ""' < $payload)
+dockerfile=$(jq -r ".params.dockerfile // \"${build}/Dockerfile\"" < $payload)
 
 load_file=$(jq -r '.params.load_file // ""' < $payload)
 load_repository=$(jq -r '.params.load_repository // ""' < $payload)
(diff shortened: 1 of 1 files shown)


assets/out
```diff
@@ -74,7 +74,7 @@ load_base=$(jq -r '.params.load_base // ""' < $payload)
 build=$(jq -r '.params.build // ""' < $payload)
 cache=$(jq -r '.params.cache' < $payload)
 cache_tag=$(jq -r ".params.cache_tag // \"${tag_name}\"" < $payload)
-dockerfile=$(jq -r '.params.dockerfile // ""' < $payload)
+dockerfile=$(jq -r ".params.dockerfile // \"${build}/Dockerfile\"" < $payload)
```
donaldguy: I can't think of a reason this is a bad idea (I tested double-slash cases without issue), but let me know

donaldguy: It may be worthwhile however to in general introduce a

```bash
if [ ! -f $dockerfile ]; then
   echo "Dockerfile doesn't appear to exist: $dockerfile"
fi
``` 

cause if it doesn't (e.g. if/when I used the `dockerfile:` key but specified the parent directory out of habit) the `grep` I added caused the process to silently fail

assets/out
```diff
@@ -89,6 +89,14 @@ if [ -n "$load" ]; then
   docker load -i "${load}/image"
   docker tag $(cat "${load}/image-id") "${repository}:${tag_name}"
 elif [ -n "$build" ]; then
+  ECR_REGISTRY_PATTERN='[0-9]\+\.dkr.ecr.\+amazonaws\.com'
```
vito: This should probably more closely match:

https://github.com/awslabs/amazon-ecr-credential-helper/blob/407daf548f87666c4f6cbefe79bf521f9a7f626f/ecr-login/ecr.go#L27

In particular it looks like the subdomain can be non-numeric.

assets/out
vito: Hmm seems fine.

willejs:
@vito any chance of a review on this?

vito:
Sorry for the delay on this. I've been hesitating lately with PRs to this resource as the shell scripting bits are getting more and more complicated and the risk of things breaking keeps growing (and in fact it's already broken in the past).

@donaldguy Looks like this needs to be rebased - and just to make me worry less can you confirm that both ECR and non-ECR repositories work?

/cc @zachgersh who I've expressed these concerns to in the past and may champion a Go rewrite :P

jamesthimont:
@donaldguy I'm keen to get this working.  Will you have a chance to update the PR?

donaldguy:
Rebased.

pivotal-issuemaster:
@donaldguy Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=66) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

donaldguy:
cla-no-sync
😞

pivotal-issuemaster:
@donaldguy Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=66)!

vito:
thanks!

PR #77, opened 2016-11-01
Changes skip_download functionality.

charlieoleary:
## Problem

Concourse automatically implements a `get` task after a `push` task to get updated version information.  Generally this is not an issue, but with Docker builds this results in a `docker pull` being performed.  In many cases this can lead to significant time and bandwidth consumption on builds.  The docker-image resource implements a `skip_download` parameter, but this only applies to manually created `get` tasks and there is no way to apply a `skip_download: true` to the `get` action that is automatically performed.

## Solution

My proposed solution is to change the `skip_download` parameter to `download` and set the default action to false.  This will cause the automatic `get` action to skip the download and only get the necessary version data.  I have also added a conditional to perform the download action if the `save` parameter is set to true, as a download will be required in order to perform the save.

## Causes for Concern

The only real downside of this change is that resources that currently implement a `get` action will now have to have a `download: true` added to their parameters.  However, I would argue that the added benefits (namely potentially saving a TON of time and bandwidth for everyone) outweigh the cons of having to add this parameter.

Another concern is how it will affect tasks via the docker-image resource.  However, provided the task resource sets`save: true` it should still work as intended.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,11 +83,11 @@ The following files will be placed in the destination:
 
 #### Parameters
 
-* `save`: *Optional.* Place a `docker save`d image in the destination.
 * `rootfs`: *Optional.* Place a `.tar` file of the image in the destination.
-* `skip_download`: *Optional.* Skip `docker pull` of image. Only `/image-id`,
+* `download`: *Optional.* Default `false`. Performs `docker pull` of image. If not set to true, Only `/image-id`,
   `/repository`, and `/tag` will be populated. `/image` and `/rootfs.tar` will
   not be present.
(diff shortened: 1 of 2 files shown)


vito:
This will break `image_resource` (and just about anyone using the resource today).

You actually can disable the download on `put`s, like so:

```yaml
- put: x
  get_params: {skip_download: true}
```

We've however forgotten to document `get_params`.

charlieoleary:
Ah, got it.  Then I'll document `get_params` instead -- extremely useful!

PR #80, opened 2016-11-17
Added timeout to pingClient.

cristian-radu:
We should add a timeout in case pingClient fails, otherwise check will just hang and we will never see the error that caused pingClient to fail.

### DIFF
--- a/cmd/check/main.go
+++ b/cmd/check/main.go
@@ -182,8 +182,11 @@ func makeTransport(logger lager.Logger, request CheckRequest, registryHost strin
 
 	authTransport := transport.NewTransport(baseTransport)
 
+	timeout := time.Duration(10 * time.Second)
+
 	pingClient := &http.Client{
 		Transport: retryRoundTripper(logger, authTransport),
+		Timeout: timeout,
 	}
 
 	challengeManager := auth.NewSimpleChallengeManager()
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@cristian-radu Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=80) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

pivotal-issuemaster:
@cristian-radu Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=80)!

vito:
@cristian-radu Sorry for the delay on this :/ Holiday season and all.

I'm super wary of timeouts but I think this makes sense, though 10 seconds may be a little low. How about 60 seconds since that's the default check interval? Ideally 10 seconds would be enough, but I'm always afraid to have things just never work for folks on bad networks.

tredfield:
any way to make timeout configurable?

chendrix:
I'm worried about adding a [knob](https://github.com/concourse/concourse/wiki/Anti-Patterns#knobs) and I'm also worried about this being [contributor-burden](https://github.com/concourse/concourse/wiki/Anti-Patterns#contributor-burden) since the idea of network flakiness is one that's true across all resources with external dependencies.

PR #83, opened 2016-11-29
Set docker daemon --mtu flag

### LINKED ISSUE #67: Garden MTU is ignored when starting docker daemon
As discussed in https://github.com/concourse/concourse/issues/551 garden can be configured to use a specific interface MTU.

However docker-image-resource ignores that MTU. This leads to connection problems on certain networks, while building docker images. 

[Hard coding the MTU to match our network](https://github.com/manno/docker-image-resource/commit/288fb29bfd1876e724055ba252d13ee5a778e43d)  stops zypper, called from inside the Dockerfile, from freezing.

databus23:
This lets the docker daemon adopt the mtu set on the outer gardian container via `--garden-mtu=` flag.

Fixes #67

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -49,7 +49,8 @@ start_docker() {
     mount -o remount,rw /proc/sys
   fi
 
-  local server_args=""
+  local mtu=$(cat /sys/class/net/$(ip route get 8.8.8.8|awk '{ print $5 }')/mtu)
+  local server_args="--mtu ${mtu}"
   local registry=""
 
   for registry in $1; do


assets/common.sh
```diff
@@ -49,7 +49,8 @@ start_docker() {
     mount -o remount,rw /proc/sys
   fi
 
-  local server_args=""
+  local mtu=$(cat /sys/class/net/$(ip route get 8.8.8.8|awk '{ print $5 }')/mtu)
+  local server_args="--mtu ${mtu}
```
vito: missing a " here

databus23: Oh, man. Stupid copy&paste error

vito:
thanks!

PR #87, opened 2016-12-23
Add if_not_exits param; prevents building image that already exists

ggeorgiev:


assets/out
```diff
@@ -131,7 +131,18 @@ elif [ -n "$build" ]; then
     expanded_build_args=${expanded_build_args}" "$(jq -r 'with_entries(.key |= "--build-arg " + . )|with_entries(.key = .key + "=" +.value)|keys|join(" ")' <$build_args_file)
   fi
 
-  docker build -t "${repository}:${tag_name}" $expanded_build_args -f "$dockerfile" "$build"
+  already_built=false
+  if [ "$if_not_exists" = "true" ]; then
```
jtarchie: The `if_not_exists` was never assigned/parsed from the JSON payload, so it will always be undefined. Right

jtarchie:
@ggeorgiev, we've attempted to write [tests](https://github.com/concourse/docker-image-resource/blob/master/tests/out_test.go). If you can backfill tests for your feature that be would be great!

assets/out
ggeorgiev: I am not sure what you mean with this. The change as is was working for me on 2.x.x. I did not succeed to make it work on 3.x.x because of some privilege issues. If something changed recently this code does not handle it.

ggeorgiev:
I will try to find time but I cannot promise. As I remember creating an appropriate environment to be able to develop accordingly took me a lot of time. I finished simply using the resource to test it.

assets/out
ebabani: Where is `$if_not_exists` first set? It looks like the first use is in that `if` statement.

ggeorgiev: sorry, it seems it got lost with some of the merges. I restored the initialization

PR #99, opened 2017-02-13
Implement cache using docker's --cache-from flag

databus23:
With docker 1.13 a new build flag was added: --cache-from
This flag deprecated the buildcache tool that we used with docker 1.12

Details: https://github.com/docker/docker/pull/26839


databus23:
rebased on master.
This removes a lot of cruft from the image and even makes the caching use-case a tad faster. I would really like to see this in 2.8.0.
ping @vito

vito:
sweet! thanks.

PR #113, opened 2017-04-05
Fix an incorrect error message

superbrothers:
This commit fixes an incorrect error message when the specified Dockerfile is not a file.


vito:
looks correct to me as-is; "it doesn't appear it is not a file" means "it appears it is a file" :)

superbrothers:
oh... I'm sorry.

vito:
No worries!

PR #124, opened 2017-06-22
Support for pushing multiple tags

### LINKED ISSUE #98: Support for pushing multiple tags other than 'latest'
I am looking to build and push docker images into multiple tags, none of which are 'latest'. This is to be able to push tags like `onbuild` and `onbuild-${githash}`, and then be able to use these by either taking the latest `onbuild`, or pinning the version by using the hash.

With regular images, I'm able to push both `latest` and `${githash}` tags from the same build thanks to `tag_as_latest`, but this is not possible when not using the `latest` tag for one of those (without building twice)

jerith:
Fixes #98.

This is a very quick implementation that adds a new `additional_tags` parameter. `additional_tags` should reference a file that contains a whitespace-separated list of tags.

I've done minimal testing so far, except to confirm that it does indeed push the image with each of the additional tags and that it doesn't break if `additional_tags` isn't provided.

### DIFF
--- a/assets/out
+++ b/assets/out
@@ -51,6 +51,7 @@ log_in "$username" "$password" "$registry"
 tag_source=$(jq -r '.source.tag // "latest"' < $payload)
 tag_params=$(jq -r '.params.tag // ""' < $payload)
 tag_prefix=$(jq -r '.params.tag_prefix // ""' < $payload)
+additional_tags=$(jq -r '.params.additional_tags // ""' < $payload)
 need_tag_as_latest=$(jq -r '.params.tag_as_latest // "false"' < $payload)
 build_args=$(jq -r '.params.build_args // []' < $payload)
 build_args_file=$(jq -r '.params.build_args_file // ""' < $payload)
@@ -65,6 +66,15 @@ if [ -n "$tag_params" ]; then
   tag_name="${tag_prefix}$(cat $tag_params)"
 else
   tag_name="$tag_source"
+fi
+
+additional_tag_names=""
+if [ -n "$additional_tags" ]; then
+  if [ ! -f "$additional_tags" ]; then
+    echo "additional tags file '$additional_tags' does not exist"
+    exit 1
+  fi
+  additional_tag_names="$(cat $additional_tags)"
 fi
 
 if [ -z "$repository" ]; then
@@ -178,6 +188,14 @@ if [ "$need_tag_as_latest" = "true" ] && [ "${tag_name}" != "latest"  ]    ; the
   docker tag "${repository}:${tag_name}" "${repository}:latest"
   docker push "${repository}:latest"
   echo "${repository}:${tag_name} tagged as latest"
+fi
+
+if [ -n "$additional_tag_names" ]    ; then
+  for additional_tag in $additional_tag_names; do
+    docker tag "${repository}:${tag_name}" "${repository}:${additional_tag}"
+    docker push "${repository}:${additional_tag}"
+    echo "${repository}:${tag_name} tagged as ${additional_tag}"
+  done
 fi
 
 jq -n "{


pivotal-issuemaster:
@jerith Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=124)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=124) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@jerith Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=124)!

jerith:
I haven't yet updated the docs in `README.md`. I couldn't find any automated tests, but I'm happy to add some test cases for this work if there are any that I missed.

I have been publishing my experimental images to https://hub.docker.com/r/jerithorg/docker-image-resource/ as I go, so that *might* be available for people to try out. I make no guarantees that it will be functional (or even still exist) any time after I submit this comment, so **don't use it for any builds you care about**. :-)

jtarchie:
@jerith, we've attempted to write [tests](https://github.com/concourse/docker-image-resource/blob/master/tests/out_test.go). If you can backfill the your feature that be would be great!

jerith:
I can give it a try, but it would be much easier if there were some existing tests for `tag_latest` or something that I could work from. :-)

jtarchie:
@jerith, does it make sense that this is provided as an additional option? I don't have the answer for this, just asking, how about `tag` just does it automagically?

ebabani:
@jtarchie `tag` currently takes in a path to a file. How do you see is supporting multiple tags?

If anything this would be a nice replacement of `tag_as_latest`

One really nice use case for this is the ability to tag images with a semver, as well as the branch name, so you would have access to latest `master` or `develop` images.

william-tran:
@ebabani, I think @jtarchie was suggesting that we keep `tag` to point to a path to a file, but the specification for that file can be one or more whitespace-separated tags. I like this. 

While our use case is having one variable tag (a semver), and one constant tag (a branch name), I think a file with multiple tags is the most flexible approach, and just sticking with `tag` requires the least API change.

william-tran:
Actually... just sticking with `tag` has its own complications: does `tag_prefix` apply to the first one in that file of all of them?

jerith:
I added a new parameter because I didn't want to change the behaviour of any existing parameters. I'm happy to use whitespace-separated tags in the existing `tag` file, but then it might make more sense to call it `tags` instead and deprecate `tag`.

jtarchie:
@jerith, this all makes sense!

Can I request two things:

* Update the `README`
* Are you able to provide any tests? Happy to organize a time to pair on it if you'd like.

philicious:
I'll start using this PR in the next days. 

I'm fine with having `additional_tags`, however imho it would also be fine to extend the `tag` param so it can either supply a single tag like now, or work with multiple tags. This wouldnt break old behavior. The naming `tag` instead of `tags` is rather cosmetic. One could also think of reading the "deprecated" `tag` and add it to the read list of `tags` internally.

philicious:
can someone hint me how I would build this? 
it does `ADD docker/ /usr/local/bin/` but docker dir is not part of this repo, so how should I init build env ? 
for using it, I would define a new `resource_type` "docker-image" and point it to my private registry ? or do I have to name it differently as "docker-image" is a built-in resource ?

philicious:
I was meanwhile able to build the image by putting https://get.docker.com/builds/Linux/x86_64/docker-17.05.0-ce.tgz as folder "docker" into the build dir.
however when using the resource as 
```
resource_types:
- name: docker-image-multi-tag
  type: docker-image
  privileged: true
  source:
    repository: eu.gcr.io/foobar/docker-image-resource
    username: _json_key
    password: {{gcr-key}}    
    tag: latest
```
~~I get an error when the resource is being executed:~ fixed by adding `privileged: true` to the resource_type
```
using version of resource found in cache
mount: permission denied (are you root?)
```

multi-tagging is working fine now but the subsequent `get` task is successfully pulling the image but then errors out with 
```
container_linux.go:247: starting container process caused "exec: \"/tmp/print-metadata\": permission denied"
docker: Error response from daemon: oci runtime error: container_linux.go:247: starting container process caused "exec: \"/tmp/print-metadata\": permission denied".
```
I'm not sure if this is caused by the PR or another missing configuration in my piepline

vito:
@jerith Sorry there aren't many tests to use as a starting point, but we do have [these ones](https://github.com/concourse/docker-image-resource/blob/master/tests/out_test.go) which are the pattern to follow. We're trying to backfill these and clean up some of the code debt in this resource, which is why we're hesitant to pull in PRs without them.

Based on the code change, I think the test may actually be kinda straightforward; it could just look for multiple `docker push` commands with the given tags.

vito:
@jerith Would you be able to add tests for this? There should be a pretty decent pattern set now. See https://github.com/concourse/docker-image-resource#development

gcmalloc:
Following the merge of #171, this can now be closed I think.

vito:
yep!

chaoyangnz:
If I add `additional_tags` when I push docker to ECR. And another job use this docker image resource as input, so I `get` it, how can I get the updated image tag I just specified in `additional_tags`?
as `IN`, I only read `latest` from `/tag` file. The documentation says:

> /tag: The tag of the repository that was fetched.

philicious:
@chaoyangnz the documentation is a bit misleading for `IN`
if you check the source at https://github.com/concourse/docker-image-resource/blob/master/assets/in#L32 , you see it will only read the `tag` you supplied as argument and will write that out as `/tag` file.
In https://github.com/concourse/docker-image-resource/blob/master/assets/in#L78 writes alot meta info about the image to `/docker_inspect.json`.
However, the json you get with `docker inspect` will only contain the tags you pulled before and not all available in the repo. 
For getting all available tags, you would need to query the registries API like
```
curl 'https://registry.hub.docker.com/v2/repositories/library/debian/tags/'|jq '."results"[]["name"]'
```
This could be added to the docker-image-resource tho so it would write another file with all available tags.

PR #126, opened 2017-06-30
added additional login

johnnoone:
Allow to access to more private registries. 
This is useful when you have to deal with multiple private registries, like one in hub.docker and another in grc.io

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,16 @@ Note: docker registry must be [v2](https://docs.docker.com/registry/spec/api/).
 
 * `password`: *Optional.* The password to use when authenticating.
 
+* `login`: *Optional.* An array of additional username, password and registry
+  to log in, following this format:
+
+  ```yaml
+  login:
+  - username: USERNAME
+    password: PASSWORD
+    registry: REGISTRY # (optional)
+  ```
+
 * `aws_access_key_id`: *Optional.* AWS access key to use for acquiring ECR
   credentials.
 
@@ -137,7 +147,7 @@ version is the image's digest.
   first pull `image:tag` from the Docker registry (so as to use cached
   intermediate images when building). This will cause the resource to fail
   if it is set to `true` and the image does not exist yet.
-  
+
 * `cache_tag`: *Optional.* Default `tag`. The specific tag to pull before
   building when `cache` parameter is set. Instead of pulling the same tag
   that's going to be built, this allows picking a different tag like
@@ -174,7 +184,7 @@ version is the image's digest.
   be tagged as `latest` in addition to whatever other tag was specified.
 
 * `build_args`: *Optional.*  A map of Docker build arguments.
-  
+
   Example:
 
   ```yaml
@@ -183,7 +193,7 @@ version is the image's digest.
     how_many_things: 2
     email: me@yopmail.com
   ```
-    
+
 * `build_args_file`: *Optional.* Path to a JSON file containing Docker build
   arguments.
 
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -97,6 +97,19 @@ log_in() {
   fi
 }
 
+multi_login() {
+  local credentials="$(echo $1 | jq -r '. as $data | [$data] | flatten')"
+  local credentials_count="$(echo $credentials | jq -r '. | length')"
+
+  for i in $(seq 0 $(expr "$credentials_count" - 1));
+  do
+    username="$(echo $credentials | jq -r .[$i].username)"
+    password="$(echo $credentials | jq -r .[$i].password)"
+    registry="$(echo $credentials | jq -r .[$i].registry // "")"
+    docker login -u "${username}" -p "${password}" ${registry}
+  done
+}
+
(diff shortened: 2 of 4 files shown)


pivotal-issuemaster:
@johnnoone Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=126)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=126) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@johnnoone Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=126)!

jtarchie:
@johnnoone, what is the use case for this with a single image? When you define a resource for docker it should be only 1-to-1, pull from the same registry push to the same registry. If we have multiple things could be indeterministic on what is the `latest` version being pulled.

chendrix:
I believe the (a?) use case is particularly if your `FROM` is a different registry.

johnnoone:
yes @chendrix this my use case.
Image is based from a private registry A, then is pushed to a private registry B.

jtarchie:
@johnnoone, makes sense. Are you able to provide a test for this feature?

vito:
Tests should now be fairly easy to add under `tests/`. See `out_test.go` for reference. Could you add some?

I'd also like to point towards `load_base`, which is another way of having an authenticated `FROM`.

EugenMayer:
`load_base` needs a `save` before, which makes the whole build 10x slower - using a registry with push/pull in a  DC speeds up enormously, not to speak about how resource-blocking a save is, most probably nothing else can happen during that.

vito:
@EugenMayer That's not necessarily true at all. `get` steps are cached, so it'll only need to fetch and `save` once per worker. There will be a penalty for namespacing the image, but that's something we can fix in the future, and you can even skip that if you're using the `btrfs` driver.

If you're not using `load_base` and you're fetching an from the registry on *every single build*, you lose out on proper dependency version management, and it could easily end up being slower and more error prone.

This is why I recommend features and correctness first, and *performance second*. Your pipeline is doing your work for you, there's no need to babysit it for that grueling 10 minutes. Correctness always wins. Then, with correctness, we can implement airtight optimizations when it matters, as we have in the past. Your pipelines will be correct and suddenly they'll be faster.

EugenMayer:
@johnnoone thanks, that patch actually fixed our issue the same way. Base image from A, pushing into reg B, while the login will be the for B, thus pulling A will fail during the build.

This cannot be fixed using a `get` of A in the resources, since those are simply different docker-engines and do not share the storage at all.

`save` and `load` are just here to overcome the inability of the worker to share its own storage with the actual docker-in-docker self implementation of docker-image-resource. Sharing this would enable to have far fewer pulls while using the SHA of the layer (version) one want to build on, is the version reference we would pass to docker to build upon. So its still version safe, and works properly as a fallback. Docker would pull the image if the layer is missing automatically, not need to be explicit here - if the layer is present, it just uses it ( and its correct ). But that drags this issue offtopic and starts a discussion which is waste of energy anyway. We will implement this dock `docker:did`, not need to do 99% of the stuff already present in 2017 and done by docker itself.

vito:
@EugenMayer Doing a `get` doesn't require them to run the same Docker engine or share the same storage. It's also providing versioning at a *pipeline level* - not just "version safe". It lets you know what version the build ran with, so you can see more easily if it broke things. It's just more natural in the Concourse world.

In fact, the way the feature is expressed, you shouldn't care whether it runs on the same engine, and it's possible that it could some day. That's the point of having abstractions, and thinking at a higher level.

Please don't derail other issues (**especially not PRs**) by injecting your own grievances. If you have something constructive to suggest, feel free to open a new issue.

README.md
```diff
@@ -18,6 +18,16 @@ Note: docker registry must be [v2](https://docs.docker.com/registry/spec/api/).
 
 * `password`: *Optional.* The password to use when authenticating.
 
+* `login`: *Optional.* An array of additional username, password and registry
```
vito: Can we call this `additional_login` or something to prevent confusion with `username` and `password`?

tanner-bruce:
@vito I added a new PR renaming that field, however I am unsure of where to add tests and what exactly you envision the tests doing. Since the `extra_logins` would never be used in the check, nor push, only as a side effect while doing the actual docker build I am not really sure what there is to even.

A cleaner/more testable approach in this case may be to deprecate `username` and `password`, and move to only using 

```
logins:
- username:
   password:
   registry:
```

for all basic username/password logins. Thoughts?

vito:
Closing due to inactivity. Partly my fault for not being responsive enough. If someone wants to try this again, feel free, but please include tests! The existing suite stubs out the `docker` CLI so it shouldn't be too tricky to assert that additional `login` commands are run if the new config is set.

Issue #143, opened 2017-09-03
`daemon.json` interface for resource


drahnr:
This allows to set parameters of the docker daemon i.e.
```
{     "max-concurrent-downloads": 1,     "max-concurrent-uploads": 1 }
```
to prevent timeouts due to thread starvation on a limited bandwidth up/downlink.

vito:
#144 is now merged; closing. thanks!

PR #144, opened 2017-09-03
Add concurrent upload and downloads source parameters

drahnr:
Minimal changes to address the root issue of https://github.com/concourse/docker-image-resource/issues/143


pivotal-issuemaster:
@drahnr Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=144)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=144) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@drahnr Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=144)!

assets/in
```diff
@@ -32,6 +32,8 @@ repository="$(jq -r '.source.repository // ""' < $payload)"
 tag="$(jq -r '.source.tag // "latest"' < $payload)"
 ca_certs=$(jq -r '.source.ca_certs // []' < $payload)
 client_certs=$(jq -r '.source.client_certs // []' < $payload)
+max_concurrent_downloads=$(jq -r '.source.max_concurrent_downloads // 3' < $payload || echo "3")
+max_concurrent_uploads=$(jq -r '.source.max_concurrent_uploads // 3' < $payload || echo "3")
```
jtarchie: Why the `echo 3`?

drahnr: Default value 3

jtarchie: Indeed it [does](https://stedolan.github.io/jq/manual/#Alternativeoperator://).

drahnr:
I removed it, anything else necessary?

assets/out
```diff
@@ -33,6 +33,8 @@ password=$(jq -r '.source.password // ""' < $payload)
 repository=$(jq -r '.source.repository // ""' < $payload)
 ca_certs=$(jq -r '.source.ca_certs // []' < $payload)
 client_certs=$(jq -r '.source.client_certs // []' < $payload)
+max_concurrent_downloads=$(jq -r '.source.max_concurrent_downloads // 3' < $payload || echo "3")
```
jtarchie: Can you do the same as above?

jtarchie: Is three the default the docker always provides? This is backwards change that would make other's pipelines go slower until they updated to max a higher max. Maybe we should make this optional, as in, you provide a value use it, no value, don't set any min/max.

drahnr: Yes, docker default for download is 3, upload is - for no obvious reason - defaulting to 5

vito: Is the `|| echo "3"` necessary? The `// 3` should default the value to `3` if it's not present.

drahnr: No it is not but it has been removed in https://github.com/concourse/docker-image-resource/pull/144/commits/99994f0fc18f7d7e72f713013df4034e41e30c1b

drahnr:
This is a trivial change, so can we please move forward or let me know what is necessary to change in order to get it merged or drop it altogether :)

vito:
@drahnr Hi, sorry for the wait on this - there's a lot of tech debt in this resource so we've been hesitant to process PRs without tests, and most of them don't have any since part of the debt is that there *aren't any* to reference when submitting a PR. :)

We've just finished backfilling some around the `dockerd` command with [this commit](https://github.com/concourse/docker-image-resource/commit/5edc3b8f587ff028c7db887841865b17037b4143) - could you try using it as a reference for adding tests for this PR?

p.s. the `|| echo "3"` was indeed removed from `/in` but not `/out`. 😄

drahnr:
Oops I totally missed that. Thanks! 

Will add a few tests next weekend.

drahnr:
I totally forgot about this, should be done with 42ab26f.

drahnr:
I totally forgot about this, 35e9968 should complete it.

vito:
@drahnr No worries! Another *poke* for tests and a rebase? :) (See https://github.com/concourse/docker-image-resource/blob/master/tests/out_test.go for where to add them now.)

drahnr:
@vito hopefully this suffices now, I also squashed my commits into one

vito:
@drahnr Hmm I think something's a bit off with the merge/squash/rebase/whatever. There are a bunch of commits rather than just the one. No big deal though, we can squash + merge when the time comes.

The build failed but it's our fault; fixing it now. I'll re-run it when things are stable.

drahnr:
I did just a regular rebase, but I guess the merge conflicts (for whatever reason there were any)  caused this.

assets/common.sh
```diff
@@ -57,12 +57,14 @@ start_docker() {
...
 
-  for registry in $1; do
+  server_args="${server_args} --max-concurrent-downloads=$1 --max-concurrent-uploads=$2"
+
+  for registry in $3; do
     server_args="${server_args} --insecure-registry ${registry}"
   done
 
-  if [ -n "$2" ]; then
-    server_args="${server_args} --registry-mirror $2"
```
vito: Sadly the '=' here causes the build failure. Could you replace it with a space?

vito:
Sweet, thanks!

PR #154, opened 2017-10-20
add dry_run

pickledrick:
Allows a param to be set to the out operation dry_run=true. Helps for CI testing allowing us to do dry run builds for docker images


pivotal-issuemaster:
@thebeefcake Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=154)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=154) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@thebeefcake Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=154)!

pickledrick:
@vito I presume the tests are failing as it can't do a pull on the resource that was created from the out operation. I can't really log in to view this is the case.

vito:
@pickledrick Ah it actually errored since our CI instance was deployed while it was in-flight.

Could you add tests for this? We've started laying the groundwork for doing so under `tests/out_test.go` - hopefully one of those is good enough for reference.

Thanks and sorry for the wait!

tests/out_test.go
```diff
@@ -118,5 +118,35 @@ var _ = Describe("Out", func() {
 
 			Expect(session.Err).To(gbytes.Say(docker("pull 123123.dkr.ecr.us-west-2.amazonaws.com:443/testing")))
 		})
+
+		Context("When using ECR with dry_run", func() {
```
vito: This behavior isn't related to ECR; I wouldn't put it under that context. Can it be tested without it?

tests/out_test.go
```diff
@@ -118,5 +118,35 @@ var _ = Describe("Out", func() {
...
+					"source": map[string]interface{}{
+						"repository": "test",
+					},
+					"params": map[string]interface{}{
+						"build":      "/docker-image-resource/tests/fixtures/ecr",
+						"dockerfile": "/docker-image-resource/tests/fixtures/ecr/Dockerfile",
+						"dry_run":    "true",
+					},
+				})
+				Expect(session.Out).To(gbytes.Say(docker("Successfully built")))
```
vito: Can we have the test assert that it didn't run `docker push`?

tests/out_test.go
```diff
@@ -118,5 +118,35 @@ var _ = Describe("Out", func() {
...
+			})
+
+			It("successfully builds the image multi build docker file locally", func() {
```
vito: Don't think there's any need for an additional example, since the delta in behavior should not affect this case in particular.

vito:
consolidating into #185

PR #155, opened 2017-11-05
Update README.md

EugenMayer:
Fix outdated port notice from private repos. they are miss leading and partially even wrong - one has to use the exact same notation as on the cli, or the images will be treated as different ones.

`myrepo.tld:443/image` is not `myrepo.tld/image` so if one has used ` repository: myrepo.tld:443/image` and the used `FROM myrepo.tld/image` in the `Dockerfile` during the build, it will fail with "cannot pull `myrepo.tld/image` without credentials

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ Note: docker registry must be [v2](https://docs.docker.com/registry/spec/api/).
 * `repository`: *Required.* The name of the repository, e.g.
 `concourse/docker-image-resource`.
 
-  Note: When configuring a private registry, you must include the port
-  (e.g. :443 or :5000) even though the docker CLI does not require it.
+  Note: If you have a private repository,b e sure to use the same notation as you use in `FROM` - if needed, including your custom port like :5000. You should not include :443 usually, since that is the default - but if you used `FROM myrepo.tld:443/image` in the `Dockerfile`, it has to be `repository: myrepo.tld:443/image` here. In normal cases though, it would be `myrepo.tld/image`
+
 
 * `tag`: *Optional.* The tag to track. Defaults to `latest`.
 


pivotal-issuemaster:
@EugenMayer Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=155)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=155) to manually synchronize the status of this Pull Request.

vito:
Maybe we can just remove the notice altogether?

It really just comes down to "use whatever notation you're already using". We could clarify specifically in `load_base` that the `FROM` has to match the repository + tag used for the reference artifact, though, since that's when it *really* matters.

EugenMayer:
As a technical term, its right to put it like "whatever you have been using for now", so be it with port, even with the optional 443 one, or with the mandatory non-default port.

I would still use a description here, since when making a mistake, it is not obvious - since the error will be a `could not fetch .. missing credentials` in a total different context - making this link is just hard sometimes, so being descriptive here makes it easier to avoid it probably

vito:
@EugenMayer Can we at least put the notice closer to where it's likely to be relevant? Right now it just seems like a large cause for alarm that shouldn't be a concern 99% of the time (now that even the original notice wasn't needed, and was in fact what led to the problem you hit in the first place).

You'll only hit that `missing credentials` error when trying to use an image with a mismatched `repository` with `load_base`, so that's where I'd put this warning. It could then just say something like "make sure the fetched image has the same exact repository and tag as mentioned in the `FROM` pragma" or something.

pivotal-issuemaster:
@EugenMayer Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=155)!

EugenMayer:
not going to work on this, closing

PR #159, opened 2017-12-05
Double quote build args to support complex values

### LINKED ISSUE #79: Multiline Build Arguments Break Docker Build/Put
We frequently inject ssh-keys into containers at build time (such as to allow rubygems or npm access to private git repos during build) using `--build-args`. When attempting to do this in Concourse, two different error conditions appear.

      - put: docker
        params:
          build: master
          tag: version/version
          build_args:
            KEY: {{ssh-key}}

Results in the following error:

    bad flag syntax: -----END

Using | notation to denote multiline results in slightly different behavior.

      - put: docker
        params:
          build: master
          tag: version/version
          build_args:
            KEY: |
              {{ssh-key}}

Error:

    "docker build" requires exactly 1 argument(s).
    See 'docker build --help'.

    Usage:  docker build [OPTIONS] PATH | URL | -

    Build an image from a Dockerfile

### LINKED ISSUE #158: Multivalue build_args does not work.
Passing build args with spaces in it does not work at. 
```
build_args:
  PHP_BUILD_ARGS: --enable-fpm --enable-cli
```
will result in: 
```
waiting for docker to come up...
Login Succeeded
unknown flag: --enable-cli
See 'docker build --help'.
```

Wrap the values in quotes in the Pipeline.yml does not work due to there are removed on `set-pipeline`

I suggest to wrap the whole build-arg in qoutes in `assets/out` lines 116 and 120, so the target output will be `--build-arg "PHP_BUILD_ARGS=--enable-fpm --enable-cli"`

jfmyers9:
This change properly escapes multiline/complex build arg values when calling `docker build`.

Fixes #158, Fixes #79 

*Note*: We couldn't get the tests to run on OSX. We think the `README.md` needs some updating?

Signed-off-by: Lyle Franklin <lfranklin@pivotal.io>

### DIFF
--- a/tests/fixtures/bin/docker
+++ b/tests/fixtures/bin/docker
@@ -1,2 +1,6 @@
 #!/bin/bash
 echo "DOCKER:" "$@"
+
+for var in "$@"; do
+    echo "DOCKER ARG:" "$var"
+done
--- a/tests/out_test.go
+++ b/tests/out_test.go
@@ -34,6 +34,10 @@ var _ = Describe("Out", func() {
 		return session
 	}
 
+	dockerarg := func(cmd string) string {
+		return "DOCKER ARG: " + cmd
+	}
+
 	docker := func(cmd string) string {
 		return "DOCKER: " + cmd
 	}
@@ -53,6 +57,31 @@ var _ = Describe("Out", func() {
 		})
 
 		Expect(session.Err).To(gbytes.Say(dockerd(`.*--data-root /scratch/docker.*`)))
+	})
+
+	Context("when build arguments are provided", func() {
+		It("passes the arguments correctly to the docker daemon", func() {
+			session := put(map[string]interface{}{
+				"source": map[string]interface{}{
+					"repository": "test",
+				},
+				"params": map[string]interface{}{
+					"build": "/docker-image-resource/tests/fixtures/build",
+					"build_args": map[string]string{
+						"arg1": "arg with space",
+						"arg2": "arg with\nnewline",
+						"arg3": "normal",
+					},
+				},
+			})
+
+			Expect(session.Err).To(gbytes.Say(dockerarg(`--build-arg`)))
+			Expect(session.Err).To(gbytes.Say(dockerarg(`arg1=arg with space`)))
+			Expect(session.Err).To(gbytes.Say(dockerarg(`--build-arg`)))
+			Expect(session.Err).To(gbytes.Say(dockerarg(`arg2=arg with\nnewline`)))
+			Expect(session.Err).To(gbytes.Say(dockerarg(`--build-arg`)))
+			Expect(session.Err).To(gbytes.Say(dockerarg(`arg3=normal`)))
+		})
 	})
 
 	Context("when configured with a insecure registries", func() {
--- a/assets/out
+++ b/assets/out
@@ -111,13 +111,23 @@ elif [ -n "$build" ]; then
     cache_from="--cache-from ${repository}:${cache_tag}"
   fi
 
-  expanded_build_args=""
+  expanded_build_args=()
   if [ -n "$build_args" ]; then
-    expanded_build_args=$(echo $build_args|jq -r 'with_entries(.key |= "--build-arg " + . )|with_entries(.key = .key + "=" +.value)|keys|join(" ")')
+    keys=($(echo "$build_args" | jq -r 'keys | join(" ")'))
+    for key in "${keys[@]}"; do
+      value=$(echo "$build_args" | jq -r --arg "k" "$key" '.[$k]')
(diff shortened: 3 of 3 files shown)


pivotal-issuemaster:
@jfmyers9 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=159)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=159) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@jfmyers9 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=159)!

jfmyers9:
We believe this change may fix the above issues. As said we struggled to run the `out` tests when running `docker build`. The intention of this PR is to hopefully drive the conversation forward on the linked issues.

ljfranklin:
Don't merge this, it doesn't work :) Fixing it now...

ljfranklin:
Updated, should be working now. This is the simplest solution we could find which correctly quoted the values of build args.

vito:
Sorry, the build has been failing for a bit and we hadn't noticed. It should be fixed now; I re-ran the build and it's green.

jfmyers9:
@vito This was somewhat difficult to test due to nature of the out integration tests and the behavior of the `echo` command.. Let me know what you think of the test case added.

tests/fixtures/bin/docker
```diff
@@ -1,2 +1,6 @@
 #!/bin/bash
 echo "DOCKER:" "$@"
+
+for var in "$@"; do
+    echo "DOCKER ARG:" "$var"
```
vito: :+1: makes sense to test it this way

tests/out_test.go
```diff
@@ -55,6 +59,31 @@ var _ = Describe("Out", func() {
...
+						"arg2": "arg with\nnewline",
+						"arg3": "normal",
+					},
+				},
+			})
+
+			Expect(session.Err).To(gbytes.Say(dockerarg(`--build-arg`)))
+			Expect(session.Err).To(gbytes.Say(dockerarg(`arg1=arg with space`)))
+			Expect(session.Err).To(gbytes.Say(dockerarg(`--build-arg`)))
+			Expect(session.Err).To(gbytes.Say(dockerarg(`arg2=arg with\nnewline`)))
```
vito: is this correct? since it's a backtick-quoted string, it'll contain literally `"\\n"`, not a linebreak, so this is asserting that it sees `"DOCKER ARG: arg2=arg with\\nnewline"`. Shouldn't that have an actual linebreak, not `\n`?

or does the docker cli interpret that character as a linebreak when parsing the build args?

jfmyers9: Good question.

It turns out both:
```
        Expect(session.Err).To(gbytes.Say(dockerarg(`arg2=arg with\nnewline`)))
```
and
```
        Expect(session.Err).To(gbytes.Say(dockerarg(`arg2=arg with
  newline`)))
```
pass the test. From investigating the output, it definitely is receiving the argument with a newline. I wonder if `gbytes.Say` is causing it to behave unexpectedly.

jfmyers9: That appears to be the case. When I change the assertion to:
```
       Expect(string(session.Err.Contents())).To(ContainSubstring(dockerarg(`arg2=arg with\nnewline`)))

```
it fails as expected, while the following passes:
```
       Expect(string(session.Err.Contents())).To(ContainSubstring(dockerarg(`arg2=arg with
 newline`)))
```

`gbytes.Say` converts the string to a regex, and `\n` matches the newline character.

jfmyers9: That being said, I think the assertion in the test is still correct.

vito:
Ah right, forgot Say does regexes. Makes sense. Thanks for checking!

vito:
Thanks!

krishicks:
I wanted to pass GOOGLE_CREDENTIALS_JSON as a build arg and so also ran into this issue. I pulled this repo, built the image, tried using it, and am seeing:

```
/opt/resource/common.sh: line 47: SKIP_PRIVILEGED: unbound variable
```

krishicks:
The fix for this was to set `privileged: true` in the `resource_type` declaration per a search in Concourse Slack.

krishicks:
Actually, I take that back. It still doesn't work for me; I had to add back the `privileged: true` only because I re-built the image having reverted the commit that adds the SKIP_PRIVILEGED switch.

krishicks:
Added https://github.com/concourse/docker-image-resource/pull/160 to fix the above and added documentation for others.

PR #160, opened 2017-12-14
Add default to SKIP_PRIVILEGED

krishicks:
In the event that SKIP_PRIVILEGED is not set and common.sh is sourced
into a file that has `set -u`, it will error:

```
/opt/resource/common.sh: line 47: SKIP_PRIVILEGED: unbound variable
```

I discovered this while trying to build HEAD of docker-image-resource to
be used as a resource_type. The image built but put was failing with the
above error.

The test used will substitute parameter if it is set and not null,
otherwise set null if the parameter is null, otherwise use "" (null) if
it is unset. See http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02 for a table explaining each.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,10 +240,28 @@ environment is consistent across any `docker` enabled platform. When the docker
 image builds, the test are run inside the docker container, on failure they
 will stop the build.
 
-Run the tests with the following command:
+Build the image and run the tests with the following command:
 
 ```sh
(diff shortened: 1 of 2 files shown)


vito:
Woops, didn't see this PR and fixed it myself. I'll resolve the conflict and merge anyway for the README change - thanks!

PR #167, opened 2018-01-04
Add support for multiple load bases

krishicks:
Docker supports [multi-stage builds](https://docs.docker.com/engine/userguide/eng-image/multistage-build/#use-multi-stage-builds), which allows for creating smaller resulting Docker images by throwing away intermediate containers.

Using multi-stage builds typically results in two or more `FROM` statements in the `Dockerfile`, so it makes sense to want to load multiple images to speed up building images.

I've added this as `load_bases` for simplicity, but I realize there's history with `load` and `load_base` as evidenced by `assets/out`.

I'm all for using a different key; when one is settled on I'll update the README as well.

Sample configuration:

```
  - put: deploy-image
    params:
      build: workspace
      dockerfile: workspace/Dockerfile
      load_bases:
      - golang
      - debian
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,9 @@ version is the image's digest.
   load` before running `docker build`. The directory must have `image`,
   `image-id`, `repository`, and `tag` present, i.e. the tree produced by `/in`.
 
+* `load_bases`: *Optional.* Same as `load_base`, but takes an array to load
+  multiple images.
+
 * `load_file`: *Optional.* A path to a file to `docker load` and then push.
   Requires `load_repository`.
 
--- a/assets/out
+++ b/assets/out
@@ -81,6 +81,7 @@ fi
 load=$(jq -r '.params.load // ""' < $payload)
 
 load_base=$(jq -r '.params.load_base // ""' < $payload)
+load_bases=$(jq -r '.params.load_bases // empty' < $payload)
 build=$(jq -r '.params.build // ""' < $payload)
 cache=$(jq -r '.params.cache' < $payload)
 cache_tag=$(jq -r ".params.cache_tag // \"${tag_name}\"" < $payload)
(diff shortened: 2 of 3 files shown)


vito:
Looks great. `load` and `load_base` are actually both valid, but they're used for different things. `load` is just "load an image and push it", while `load_base` is "load an image, build, and push the built image".

Would you mind adding some tests for this? We're trying to reduce the tech debt in this resource and have been backfilling tests and requiring them for new PRs.

See https://github.com/concourse/docker-image-resource/blob/master/tests/out_test.go - looks like it should be pretty straightforward, based on the other ones.

krishicks:
@vito Added a test. It _was_ pretty easy to do. Yay!

One interesting note is added as a comment in the test, which is that `docker load` seems to work just fine with a text file being given it; I'd prefer this to be an actual minimal real tarball that it could load, but since it works as-is I left it. :shrug:

krishicks:
I need to make an update to this; I'm not seeing the expected behavior.

krishicks:
What I'm seeing is the loaded images aren't being used:

```
Loaded image ID: sha256:f1c23f11852e220eea0a1b2ed9b6a33fbf4231046a58c33ab518bda51d4d28a3
Loaded image ID: sha256:5c4fa780951b060bb0a75355765bc58112350d9974970d60561671d552aaf2e2
Sending build context to Docker daemon  64.51kB
Step 1/8 : FROM golang:alpine as builder
alpine: Pulling from library/golang
605ce1bd3f31: Pulling fs layer
d1e3a1512603: Pulling fs layer
85c31f078f75: Pulling fs layer
...snip...
```

When I hijack the container and load them with the same commands they're used, but not when running through a normal `out`. 🤕

krishicks:
I take it back; `load_bases` is working correctly; my config was just specifying the wrong tag to be pulled, and so it didn't match the Dockerfile, and was therefore ignored.

vito:
Looks good! Thanks.

PR #172, opened 2018-02-06
fix #168 

### LINKED ISSUE #168: Getting WARNING! Using --password via the CLI is insecure. Use --password-stdin when trying to put a docker-resource
resources:

  - name: git-repo
    type: git
    source:
      branch: master
      private_key: |-
        -----BEGIN RSA PRIVATE KEY-----
        ----- Key here -------------
        -----END RSA PRIVATE KEY-----
      uri: <gitUrl>

  - name: docker-image
    type: docker-image
    source:
      password: <password>
      repository: <image name with repository>
      username: <username>

jobs:
  - name: test-docker
    plan:
      - get: git-repo
        trigger: true
      - put: docker-image
        params:
          build: git-repo/
          tag: 1.0.1

The pipeline works fine except that when it tries to put the docker-image it throws warning as 
"WARNING! Using --password via the CLI is insecure. Use --password-stdin."

Docker encourages using --password-stdin to login  to prevent the password from ending up in the shell’s history, or log-files. Can you please make the necessary changes so that we can avoid any password leaks

https://docs.docker.com/engine/reference/commandline/login/#extended-description

cappyzawa:
* I fixed #168 
* I traded in `-p ${password}` for `--password-stdin` when `docker login`.
* I added `password_file` to out's param.
  * The reason for adding `password_file` is that when users are trying to use a private registry, I thought that they might emit a password with another task.
* If `password_file` params exists, The source config `password` is overwritten.


cmd/check/main.go
```diff
@@ -191,7 +191,7 @@ func makeTransport(logger lager.Logger, request CheckRequest, registryHost strin
 
 	pingClient := &http.Client{
 		Transport: retryRoundTripper(logger, authTransport),
-		Timeout: 1 * time.Minute,
+		Timeout:   1 * time.Minute,
```
cappyzawa: result of executing `go fmt`

vito:
Thanks for submitting this! Just noticed that warning myself.

I think we can do without the extra param and just `echo` the password to the CLI. The param sounds like an example of https://github.com/concourse/concourse/issues/684

cappyzawa:
@vito Thanks for telling me an example.
I decided not to add a parameter.

I fixed based on your advice! 

But CI is failed.  
https://ci.concourse-ci.org/teams/resources/pipelines/docker-image/jobs/prs/builds/87

I don't change Dockerfile...

cappyzawa:
The above was an error when fetching http://dl-cdn.alpinelinux.org/alpine/edge/community/x86_64/APKINDEX.tar.gz

The error seems to be solved.

vito:
Sorry, ended up fixing this myself while involved with other things. Thanks for going through the trouble to submit a PR though!

PR #177, opened 2018-02-28
docker: force create systemd cgroup

### LINKED ISSUE #174: docker-image-resource fails on cgroups error with alpine:edge
Rebuilding the docker-image-resource locally causes all builds to fail with `cgroups: cannot found cgroup mount destination: unknown`

alpine:edge is currently running 17.12. alpine:3.7 also fails, with docker version 17.10, however alpine:3.6 works as expected. 3.6 is running docker v17.05

May be worth considering using the official docker in docker image.

Tested with a very kosher pipeline and dockerfile (from alpine:3.6; apk add bash)

```WARNING! Using --password via the CLI is insecure. Use --password-stdin.
Login Succeeded
Sending build context to Docker daemon  102.9MB
Step 1/2 : FROM alpine:edge
edge: Pulling from library/alpine
1eae7a7426b0: Pulling fs layer
1eae7a7426b0: Verifying Checksum
1eae7a7426b0: Download complete
1eae7a7426b0: Pull complete
Digest: sha256:79c2c5f6db53da44f90bb2731f29f725b5b14c378407a123776b6d3c76e6aebe
Status: Downloaded newer image for alpine:edge
 ---> 5c4fa780951b
Step 2/2 : RUN apk add --no-cache bash
 ---> Running in 9712894311cf
cgroups: cannot found cgroup mount destination: unknown```

mook-as:
This appears to now be necessary to start docker; otherwise we get a confusing `cgroups: cannot found cgroup mount destination: unknown` error the first time a docker step involves running executables.

Fixes #174

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -39,6 +39,11 @@ sanitize_cgroups() {
       ln -s "$mountpoint" "/sys/fs/cgroup/$sys"
     fi
   done
+
+  if ! test -e /sys/fs/cgroup/systemd ; then
+    mkdir /sys/fs/cgroup/systemd
+    mount -t cgroup -o none,name=systemd none /sys/fs/cgroup/systemd
+  fi
 }
 
 start_docker() {


jamespamplin:
I tested this fix with a job of mine failing due to #174, and unfortunately I couldn't get it to work. Only thing I could get working is to use `alpine:3.7` instead - see my comment on #174

vito:
Thanks, we'll try this out and see if it fixes our pipeline!

vito:
Looks good to me. The `bin-smoke` job in our pipeline would reliably fail before, but passes with this change. Thanks again!

PR #186, opened 2018-03-27
Add support for specifying a target build stage for multi-stage builds

irfanhabib:
This adds support for building a docker image from a multi-stage Dockerfile with a specified target build stage.


ghostsquad:
@irfanhabib instead of supporting specific docker parameters, it would nice to just support arbitrary docker arguments.

vito:
@irfanhabib Thanks, looks great! Could you add tests though?

@ghostsquad That would go against our [kitchen sink config](https://github.com/concourse/concourse/wiki/Anti-Patterns#kitchen-sink-config) anti-pattern.

ghostsquad:
@vito fair.

irfanhabib:
@vito I've added tests. Thanks

Issue #148, opened 2017-09-13
Multi stages build cache work only with last stage


kwladyka:
Dockerfile
```
FROM clojure:lein-alpine as dependencies
WORKDIR /app
COPY project.clj .
RUN lein deps

FROM dependencies
WORKDIR /app
RUN lein test

FROM dependencies as builder
WORKDIR /app
RUN lein uberjar

FROM anapsix/alpine-java
EXPOSE 8080
WORKDIR /app
COPY --from=builder /app/target/*-standalone.jar ./app.jar
CMD ["java", "-cp", "/app/app.jar", "clojure.main", "-m", "etingroup.core"]
```

It looks like only this part (last step) is cached:
```
FROM anapsix/alpine-java
EXPOSE 8080
WORKDIR /app
COPY --from=builder /app/target/*-standalone.jar ./app.jar
CMD ["java", "-cp", "/app/app.jar", "clojure.main", "-m", "etingroup.core"]
```

Is it possible to improve?

chrishiestand:
This should now be possible with `cache_from`. The complication is that you have to save each intermediate image (everything up until the new `FROM`) as its own image, and include that image in the `cache_from` array. This is a docker limitation afaik.

zivori:
I see the base image cached locally as <none> : <none> but how can I tag this to push it to remote registery (so it can be used in the future for caching)
`
REPOSITORY                       TAG                 IMAGE ID            CREATED             SIZE
<none>                           <none>              7dbdf83dd6a0        8 seconds ago       690MB
node                             10.1.0-alpine       1c517a20b2fd        8 weeks ago         71MB
`

kwladyka:
BTW There is connected topic to this issue
https://github.com/docker/cli/issues/666

chrishiestand:
@zivori something like this should work:

```bash
docker tag 7dbdf83dd6a0 my-intermediate-image:latest
docker push my-intermediate-image:latest
```

dhinus:
This is a full example of how I am doing multi-stage builds using `cache_from`, hope it helps. Please note I am also using `load_bases` to make sure the base image (`node:8-alpine` in my case) is cached.

```
resources:
  - name: node
    type: docker-image
    source:    
      repository: node
      tag: "8-alpine"

  - name: docker-app
    type: docker-image
    source:
      repository: my/app
  
  - name: docker-app-stage1
    type: docker-image
    source:
      repository: my/app
      tag: stage1-latest
    
jobs:
[...]

Issue #121, opened 2017-06-08
load_cache or similar support


matiasanaya:
I want to know if it's possible to use a `get` resource instead of pulling down the image when `cache: true` is set for building.

Say this is my current job (which has the nice benefit that the `FROM` image can be cached by `concourse`):

```
- get: repo
- get: a-cached-docker-image-from
  params: { save: true }
- put: a-docker-image
      get_params: { skip_download: true }
      params:
        build: repo
        cache: true
        cache_tag: a-cached-tag
        load_base: a-cached-docker-image-from
```

I'd like for something like the following to work equivalently (where the `cache` image can be cached by `concourse`):

```
- get: repo
- get: a-cached-docker-image
  params: { save: true }
- get: a-cached-docker-image-from
  params: { save: true }
- put: a-docker-image
      get_params: { skip_download: true }
      params:
        build: repo
        load_cache: a-cached-docker-image # this does not exist
        load_base: a-cached-docker-image-from
```

jtarchie:
I'm pretty confused. What would the difference in `load_base` and `load_cache`? Wouldn't they just both be loading the layers from an image? Why do you need both?

matiasanaya:
@jtarchie did I get it wrong? The way I understand it `load_base` will be used as the `FROM ...` image while `cache: true` is using a _resulting_ image in hopes it can re-use a layer. `load_cache` would basically allow for those layers to be populated. Maybe a solution with only one attribute which takes an array would do. i.e.
```
load_layers:
  - a-cached-docker-image
  - a-cached-docker-image-from
```

jtarchie:
No, I just don't understand docker that well. There are too many layers to it. :trollface: 

Thanks for the explanation. I think there might be a way to do what you are requesting. I believe others have done it.

matiasanaya:
See https://github.com/concourse/docker-image-resource/pull/188

PR #188, opened 2018-03-30
Add cache_from parameter

dhinus:
~This is not ready to be merged, but I'd like to get some feedback on the idea.~

This pull request adds a `cache_from` parameter that translates into the `--cache-from` option of `docker build`. It lets you provide local images (loaded via `load_bases`) that Docker can use as a cache during the `docker build` command.

This has two advantages:

* it is faster than using `cache: true` because it leverages the Resource caching (i.e. the Docker image is not fetched from the remote registry on every build)
* it lets you specify multiple caches, which is very useful for multi-stage builds (you can load one image per each stage, allowing Docker to use layer caching for all stages).

This pull request somewhat addresses #148 and #121.

~Any gotchas you can think of? If it looks good, I'm happy to complete this pull request with documentation and tests.~


pivotal-issuemaster:
@dhinus Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=188)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=188) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@dhinus Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=188)!

dhinus:
Example job I'm using with this:

```
- name: web
  plan:
    - get: my-project
      trigger: true
    - get: docker-node
      params:
        save: true
    - get: docker-nginx
      params:
        save: true
    - get: docker-web-stage1
      params:
        save: true
    - get: docker-web
      params:
        save: true
    - do:
      - put: docker-web-stage1
        params:
          build: my-project/web
          load_bases:
            - docker-node
            - docker-web-stage1
          dockerfile: my-project/web/Dockerfile-stage1
        get_params:
          save: true
      - put: docker-web
        params:
          build: my-project/web
          tag_as_latest: true
          load_bases:
            - docker-node
            - docker-nginx
            - docker-web-stage1
            - docker-web
        get_params:
          skip_download: true
```

vito:
Sorry, a bit lost in the context here - is this *overloading* `load_bases` as a sort of `load_caches`? Doesn't `load_bases` already result in not needing to fetch the base images in a multi-stage dockerfile?

dhinus:
Hi Alex,

I admit I'm a little lost myself, the way Docker cache works is a bit mysterious at times. :)

> is this overloading load_bases as a sort of load_caches?

Yes. I considered having a separate attribute called `cache_from_images` or `load_caches` but then I noticed `load_bases` seems to have a similar aim and I could conflate the two. Maybe this is not the best idea, I'm very open to suggestions.

> Doesn't load_bases already result in not needing to fetch the base images in a multi-stage dockerfile?

If I understand this correctly (and I'm not 100% sure) this works for "base images", i.e. images you specify in a `FROM` statement. But since Docker 1.10 images loaded via `docker load` are no longer trusted to be used as a cache, i.e. to skip build steps.

To reuse an existing image as a cache, `load_bases` is currently useless. For single-stage Docker files, you can use `cache: true` instead, even it is quite slow as it pulls the image every time from the network. For multi-stage builds, I couldn't find a way to get two consecutive builds to reuse _all_ layers from cache, hence the change I'm proposing here. I think it can be useful even for single-stage builds, because for large images it can be considerably faster than using `cache: true`.

Below is a simplified version of my multi-stage Dockerfile.

```yaml
FROM node:6 as stage1

COPY . /app
RUN npm run build

FROM nginx

COPY --from=stage1 /app/build /usr/share/nginx/html
```

My aim is that if nothing has changed in my source code I want Docker to reuse cache for _all_ the steps, up to my final custom image, so I need to load 4 images: `node:6`, `myapp:stage1-latest`, `nginx` and `myapp:latest`.

Rethinking about it, if we added a separate `load_caches` attribute, I could probably put the two "base images" (node and nginx) in `load_bases` and the two "cache images" (myapp) in `load_caches`. I would still need to `docker load` all four, but I _think_ I only need to use `--cache-from` for the "cache images". Does it make sense? I will try this out.

Side note: to generate `myapp:stage1-latest` I currently have a separate Dockerfile `Dockerfile-stage1` that corresponds to the first part of the Dockerfile above. If #186 gets merged, I could use that switch to do a build of stage 1 only and tag it as `myapp:stage1-latest`.

vito:
Gotcha! That makes sense, thanks for the info.

Man, there are way too many knobs on this resource as it is, so I'm really trying to think about this in a way that doesn't overload existing params but also doesn't just add to the pile. I've captured some thoughts on this in https://github.com/concourse/docker-image-resource/issues/190.

@dhinus What do you think of adding a `cache_from` param instead, which would be a superset of `load_bases` and result in `--cache-from` flags passed to Docker? (I say this knowing full well it's yet another param, but I'd rather be explicit about these separate use cases so it'll be easier to understand workflows when it comes time to carve up this resource.)

dhinus:
`cache_from` sounds good to me. Does it have to be a superset of `load_bases` though? If my hypothesis is correct, using a _subset_ of `load_bases` should work. I'll test this and update this pull request.

dhinus:
@vito I have pushed a new commit, and updated the title and description of this pull request.

I have tested it on a pipeline and it works as expected. I'll add tests and documentation later this week.

dhinus:
@vito this is ready for merge AFAIC. Documentation was tricky to write concisely, let me know if you think it's not clear.

vito:
@dhinus Sorry for the delay and thanks for keeping the ball rolling!

What I meant by a superset was that it should perform the same work that `load_bases` does, and *additionally* result in a `--cache-from` for each image it loaded. I think this is better than requiring the user to specify the same image in both, since `load_bases` is originally intended just for loading the images for the `FROM` pragmas. If it can work independently, there's no inter-dependence between the two params, and we're not overloading `load_bases`.

dhinus:
Sounds reasonable, I'll push an update shortly!

vito:
Perfect, thanks!

Issue #190, opened 2018-04-02
Laundry list of smells from this resource


vito:
This resource hasn't aged well. We've long suspected a new approach may be necessary, potentially including a complete rewrite and/or breaking it up into separate resource types, each focused on a specific workflow.

I'm going to use this issue as a dumping ground of sorts for now, in hopes that it eventually leads to a more focused direction that can skewer a few of these piranhas at once.

1. There are too many knobs, and way too many ways to `put`.
1. The testing situation is improving, but pretty bad.
1. Important parts are written in `bash` and are subject to the `docker` CLI output format changing. Some things are already broken because of this.
1. There are too many ways to configure credentials.
1. There's no caching by default, and it's unclear if everyone would want that. `cache: true` even pulls down the image willy-nilly, preventing Concourse caching of the image (ironically). (see https://github.com/concourse/docker-image-resource/issues/121, https://github.com/concourse/docker-image-resource/issues/148, https://github.com/concourse/docker-image-resource/pull/188)
  * Caching all the time is great and all until it's not. Lines like `apt-get update && apt-get -y install foo` will need the cache busted in order to get a new `foo`.
1. There isn't a great way to integrate `Dockerfile`s with Concourse-fetched dependencies while also having the image easy to build locally during development (you'd be missing all the dependencies).
1. There's no way to build without pushing to a registry. To be honest, as a resource that makes sense (there must be an external source of truth), but a reusable task would be great for this.
1. It embeds a Docker daemon. Would it be easier to point at an external one? What impact might this have on caching? Would this be an acceptable assumption for most people? How would the resource be pointed at the daemon?

jchesterpivotal:
There may be value in adopting one of the alternatives that are emerging -- [buildah](https://github.com/projectatomic/buildah), [Img](https://github.com/genuinetools/img), [kaniko](https://github.com/GoogleContainerTools/kaniko) etc. There is also a nascent "Container Builder Interface" (CBI) [being spruiked](https://github.com/containerbuilding/cbi) which aims to unify these under a common API. 

We can definitely seek advice and input from folks in the Google container tooling teams (tagging @mattmoor, @dlorenc and @ImJasonH), who are attacking this stuff from a lot of angles. @sclevine is also exploring this problem in the context of buildpacks, including looking closely at the various options I linked.

vito:
[...]

Issue #73, opened 2016-09-27
Support for Google Container Registry?


josdotso:
Usually, I can use `gcloud docker -a` before running docker commands that don't understand GCR and get by, but that doesn't seem to be working for me with `concourse/lite` Vagrant box.

https://cloud.google.com/container-registry/docs/advanced-authentication

Is GCR supported?

vito:
I suppose this'll need changes similar to supporting Amazon ECR. Don't know enough about Google's registry, myself.

MaikuMori:
It already works if you follow the tutorial which you linked:

```
 - name: build-image
   type: docker-image
   source:
     repository: gcr.io/project/repo
     username: _json_key
     password: {{gcr-key}}
```

You need to use `username` set to `_json_key`. And password is the `json` key file for your service account.

josdotso:
@maikuMori, I tried this previously and got a JSON not string type error.  I'll try to reproduce and post error.   Thanks

MaikuMori:
The best to also format the JSON as single line. Can be done with something like:

```bash
cat ~/key.json | jq -rc '.' | xsel -i
```

josdotso:
Thanks again @MaikuMori .. A peer confirmed it.  I must have missed something when I looked at it.  Feel free to close.

rjain-pivotal:
this doesnt work with gcr anymore....

error getting credentials - err: exit status 1, out: `2017-12-20T06:12:39Z [ERROR] Error parsing the serverURL: docker-credential-ecr-login can only be used with Amazon EC2 Container Registry.

PR #193, opened 2018-04-22
Google Container Registry (gcr.io) support

evandbrown:
This change adds support for authenticated access to gcr.io.

Authentication is limited to VM service accounts, but specifying a service account key will be supported in the future. There's a few important things happening here:

* I moved from godep to dep because I could not for the life of me convince godep to vendor http://github.com/GoogleCloudPlatform/docker-credential-gcr/ in a way that allowed `docker-credential-gcr` to be built in the Dockerfile. I tried. I really tried.

* Moving to dep allowed remove of `scripts/deps`. Now that I think about that, maybe it should stay and just run `dep ensure`. Or the docs should be updated to talk about the (new) use of dep, assuming that change makes it through.

* This currently depends on the VM running `check`, `in`, or `out` being associated with a service account that has access to GCR. Support for specifying a service account key (json) with the right prems is next on the list.

One question: are there integration tests that verify `check`?

Thanks, and LMK what questions I can answer!


assets/common.sh
```diff
@@ -105,6 +105,13 @@ log_in() {
   else
     mkdir -p ~/.docker
     echo '{"credsStore":"ecr-login"}' >> ~/.docker/config.json
+
+    # If the image is in gcr.io, configure that credential helper
+    if grep -q '^gcr.io[^ ]*' <<< ${registry}; then
+      # This will overwrite the default ECR provider
```
evandbrown: Is this OK?

vito:
Thanks for submitting this, but I'm pretty afraid to pull in substantial changes to this resource in light of a bunch of the points raised in https://github.com/concourse/docker-image-resource/issues/190

According to https://github.com/concourse/docker-image-resource/issues/73#issuecomment-255165396 folks can use GCR with a JSON key, at least, but it's kind of hacky.

More and more registries are showing up with their own auth schemes - it's becoming a sore point of this resource to have them all baked in to one big interface. (Not the *biggest* sore point, but it doesn't help.)

Dividing by registry type has been one idea for stemming the tide of complexity here. Would there be any interest in that approach? i.e. a smaller, nimbler `gcr-image` resource. Maybe one that could even integrate more deeply with GCP? Could it use Container Builder to avoid running a daemon internally? That's one of the pain points mentioned in #190, kind of related to caching semantics too.

If there is interest, I'd like to work together and work out what the interface should look like - a carbon copy of this resource would be a mistake, since it's already far too crazy. Starting small and focused on a particular workflow would go quite far IMO. Alternatively, if we divide this resource by workflow, maybe we could work out an auth interface that's easier to extend without bloating the rest of the interface.

If you don't have time and this feels like scope creep that's fine too!

evandbrown:
Those issues you linked are helpful. It seems like a lot of the pain is in `put`; I don't think I've got a good perspective there as I've primarily used this as a `task`'s `image_resource`.

A resource-per-registry is an interesting idea in the context of Container Builder, but there's probably a lot of boilerplate we'd want to share between registry implementations. I'm inclined to think a single resource that a) standardizes on Docker's credential helper; and b) has a convention for adding new providers and their configuration would be nice. If the existing `docker-image-resource` was refactored with those things in mind, would you expect to maintain backwards compatibility?

Oh, and replacing Docker with something like kaniko would be nice, too. Let me find some time this week to play around with that and try to wrap my head around `put` a little better, and to fiddle with kaniko.

Oh, and the JSON key support for authenticating to GCR does work. We're primarily trying to allow customers to authenticate from a VM associated with a service account and avoid populating credhub/vault with keys and rotating them.

Thanks for the feedback, Alex!

jchesterpivotal:
I've definitely nagged anyone at GCP within earshot (eg @erjohnso) about my dream world where Google provides officially-supported Concourse resources for some or all of the GCP APIs, including chipping in for resources that are generic on the API.

As an aside, it would be interesting to experiment with using Google Cloud Builder step images as task images in Concourse.

evandbrown:
@jchesterpivotal I’ve got plans to make your dream come true :)

Great feedback from a customer on all of this: they point  out that essentially every published example (eg `pcf-pipelines`) use `docker-image` in task definitions. They use a private registry and every time the merge upstream examples they have to run transformation scripts to rewrite those task files. Probably worth thinking about how better multi-registry support intersects with this scenario.

vito:
@evandbrown Just realized I never answered this:

> If the existing docker-image-resource was refactored with those things in mind, would you expect to maintain backwards compatibility?

If we were to refactor this resource in-place, yeah, I think we'd have to. But I think it may be better in the long term to introduce smaller resources focused on particular workflows, so that they don't have the explosion of knobs present in this resource. If that ends up being more natural than splitting by registry, we can probably work out how to share common code around authentication. That would start us on a path towards deprecation for this resource.

chrishiestand:
I don't know if this makes anything in this PR unnecessary, but for those who are stuck on this, it's already possible to push and pull from gcr.io:

The trick is:
```yaml
username: _json_key
```

```yaml
- name: my-image
  type: docker-image
  source:
    repository: us.gcr.io/my-project/image-name
    username: _json_key
    password: ((key-us-gcr-auth-rw))
```

jchesterpivotal:
@chrishiestand's suggestion worked for me. However it required a fair amount of fiddling with permissions to get it to work, finally settling on a service account with the "Storage Admin" role. 

I don't know if this is just a general thing for GCR or if it's particular to using the JSON key, but it makes me uncomfortable. Fine for spiking, but I'd want something more locked down for production.

chrishiestand:
@jchesterpivotal Thanks for testing it. These are documented in GCP documentation.

Username/password:
https://cloud.google.com/container-registry/docs/advanced-authentication#json_key_file

IAM auth:
https://cloud.google.com/container-registry/docs/access-control

FWIW, I will be using this in production.

vito:
@evandbrown Considering there are existing workarounds and given the concerns raised earlier, I'm gonna close this for now. Thanks for submitting it though!

FYI, I've started to work on a new resource type for interacting with Docker registries: https://github.com/concourse/registry-image-resource 

I'd be interested in seeing what GCR support may look like in the new resource. It may feel much less scary since it's totally centered around registry operations and cuts out all the complexity with building.

Thanks again!

larssb:
> I've definitely nagged anyone at GCP within earshot (eg @erjohnso) about my dream world where Google provides officially-supported Concourse resources for some or all of the GCP APIs, including chipping in for resources that are generic on the API.
> 
> As an aside, it would be interesting to experiment with using Google Cloud Builder step images as task images in Concourse.

I've successfully used Google Cloud Builder container images in Concourse CI steps. E.g.:
- gcr.io/cloud-builders/npm

PR #195, opened 2018-04-27
Improve readme

dhinus:
Small improvements to the README:

- order params alphabetically
- move back an example to its right place
- clarify meaning of `build_args` using wording from the [official Docker docs](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg)

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,24 +141,26 @@ version is the image's digest.
 * `build`: *Optional.* The path of a directory containing a `Dockerfile` to
   build.
 
-* `build_args`: *Optional.*  A map of Docker build arguments.
+* `build_args`: *Optional.* A map of Docker build-time variables. These will be
+  available as environment variables during the Docker build, but will not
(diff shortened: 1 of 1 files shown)


mhuangpivotal:
@dhinus Thanks for taking the time to do this!

Can I ask why the params ordering was changed? I  am assuming the original ordering was in descending importance and may be more useful for users.

dhinus:
@mhuangpivotal was the original ordering really in descending importance? It looked a bit random to me. With a list of this length, I think it's not really feasible to order the entire list "by importance", alphabetical sounds like the most convenient option to me. But I would be ok with putting a few important options at the top. What do you think?

mhuangpivotal:
@dhinus You might be right that the ordering was arbitrary before. I don't really see any options important enough to warrant being on the top, so I will just merge this. Thanks again!

PR #199, opened 2018-05-19
create_ecr_repo: If the repository is an ECR repo check if it exists and create if it doesn't

hibbert:
This PR will do the following:-

* Check if the repository being built is an ECR repo
* If it is get the repository name without the ecr url
* Check if the repository exists in ECR
* If it doesn't create the repository

To do this I've also added the following dependencies:-

* awscli, installed via pip, so python and py-pip also installed
* aws_default_region for the source configuration to set the default region which is required when using the aws cli with ECR

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,8 @@ RUN set -e; for pkg in $(go list ./...); do \
 	done
 
 FROM alpine:edge AS resource
-RUN apk --no-cache add bash docker jq ca-certificates
+RUN apk --no-cache add bash docker jq ca-certificates python py-pip
+RUN pip install --upgrade awscli
 COPY --from=builder /assets /opt/resource
 RUN mv /opt/resource/ecr-login /usr/local/bin/docker-credential-ecr-login
 
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ Note: docker registry must be [v2](https://docs.docker.com/registry/spec/api/).
 * `aws_session_token`: *Optional.* AWS session token (assumed role) to use for acquiring ECR
   credentials.
 
+* `aws_default_region`: *Optional.* AWS region required when using aws cli for ECR.
+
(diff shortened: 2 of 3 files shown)


pivotal-issuemaster:
@hibbert Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=199)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=199) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@hibbert Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=199)!

hibbert:
It might be good to have a separate resource for each repository type, although you may run into problems if someone has a FROM set to one repository type but requires to push the resulting image to another.

The resource is lacking the ability to create ECR repos on the fly without this (or something similar) meaning it's not end to end automation for pushing images to the ECR. It also doesn't support assuming roles which is required if you are working across account - https://github.com/concourse/docker-image-resource/issues/202

ghostsquad:
ECR repositories, like Github repos, should not be created from a concourse pipeline. They should be created outside and assigned appropriate permissions. I highly suggest that this PR be closed.

This would definitely cause issues for us, as we use terraform to create/manage ECR repos, and we _don't_ want things created automatically if they don't exist. We want this to fail, as it also indicates a potential typo.

vito:
Gonna close this out - thanks for taking the time to submit it though!

FWIW I've started to break this resource up - the start of that work is here: https://github.com/concourse/registry-image-resource

I've yet to begin integrating this new resource type with ECR or GCR but I'd be interesting in seeing what that looks like going forward. With this new resource the scope is smaller and more manageable, so hopefully supporting different registries doesn't feel like so much cognitive overhead and risk (well, as long as we do it properly).

PR #204, opened 2018-06-05
Avoid indefinitely echoing while waiting for docker

kmacoskey:
While waiting for docker to come up, sometimes this can take awhile,
especially when docker will never come up and something else will
timeout. Having the echo within the loop can leave you with many, many
lines of repeated stdout. This is unecessary output that provides no
value and can cause issues when builds fail with many 10s of thousands of
outputted lines and the browser attempts to load the build output.

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -80,11 +80,12 @@ start_docker() {
 
   sleep 1
 
-  echo waiting for docker to come up...
-
-  until docker info >/dev/null 2>&1; do
-    sleep 1
-  done
+  if ! docker info >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


ascherbaum-pivotal:
Ran into the same problem, and this almost killed my browser.

vito:
Makes sense, though I think this'll be a bit of a cosmetic regression since now it'll always print that line. How about checking `docker info` first?

kmacoskey:
Is the above commit what you had in mind?

vito:
Yep! Thanks, I'll go ahead and merge.

Issue #157, opened 2017-11-21
Allow adding docker build options


aberey:
Passing `--build-arg` is currently supported, but there is no way to pass other build options such as `--squash`.

sbkg0002:
Will this be added to the roadmap?

PR #213, opened 2018-07-02
add support for passing build flags

gcmalloc:
this allow to pass flags to the docker build command such as --pull, --squash or --compress.
Addresses #157.

Sorry for the noise on #212


vito:
Sorry for the delay on this, and thanks for submitting it, but I think this falls under our [kitchen sink config](https://github.com/concourse/concourse/wiki/Anti-Patterns#kitchen-sink-config) anti-pattern and adds risk of breaking people's pipelines in the future as we refactor the resource.

If there are build flags you still need passed, it'd make more sense to add them as explicit params - it'll be much safer to support going forward.

Side note: I'm also wary of adding new features here in light of concerns raised in #190 - I think in the future we'll be heading closer to something like [Kaniko](https://github.com/GoogleContainerTools/kaniko) and concourse/rfcs#7.

Thanks again!

PR #218, opened 2018-07-12
Deprecate tag parameter in favor of tag_file

ghostsquad:


vito:
thanks!

antonysouthworth-halter:
Just wondering why this param is considered deprecated?

PR #220, opened 2018-07-12
Optimize Dockerfile

ghostsquad:
* Improve readability
* Use a symlink to save roughly 10mb of size

image


ghostsquad:
My editor is stupid, I didn't realize it was actually adding tabs.

vito:
thanks!

PR #234, opened 2018-08-20
Trim credentials before use

### LINKED ISSUE #232: Check fails on private registry
I'm running a private registry using the standard docker distribution tools.

Get and put work fine, but for some reason all resources stay orange:

screenshot from 2018-08-02 11-44-27

This is caused by the check failing:

```
resource script '/opt/resource/check []' failed: exit status 1

stderr:
failed to fetch digest: 401 Unauthorized
```

screenshot from 2018-08-02 11-45-22

Note how there are revisions present (those are pushed). Pulls also work. It just seems that check fails to pass authentication info.

There's no magic in the resource configuration:

```
    - name: ergast-mirror-docker
      type: docker-image
      source:
          repository: registry.rocketeer.be/ergast-mirror
          username: ((docker-username))
          password: ((docker-password))
```

Is this a bug or am I overlooking something?

rubenv:
Fixes #232

### DIFF
--- a/cmd/check/main.go
+++ b/cmd/check/main.go
@@ -227,7 +227,10 @@ func makeTransport(logger lager.Logger, request CheckRequest, registryHost strin
 	err := challengeManager.AddResponse(pingResp)
 	fatalIf("failed to add response to challenge manager", err)
 
-	credentialStore := dumbCredentialStore{request.Source.Username, request.Source.Password}
+	credentialStore := dumbCredentialStore{
+		username: strings.TrimSpace(request.Source.Username),
+		password: strings.TrimSpace(request.Source.Password),
+	}
 	tokenHandler := auth.NewTokenHandler(authTransport, credentialStore, repository, "pull")
 	basicHandler := auth.NewBasicHandler(credentialStore)
 	authorizer := auth.NewAuthorizer(challengeManager, tokenHandler, basicHandler)


vito:
Thanks for taking the time to submit a PR for this! Buuuuut, I'm not sure I like the implications of this change. It seems like the problem really lies in the data. How'd you end up with a linebreak in there? (Maybe `echo foo | ...`?)

I'm just worried that a) this is something that technically would apply to every property someone may want to put in a credential manager and I don't want to maintain a bunch of TrimSpaces, and b) someone may legitimately have whitespace in their password (who knows, it could be generated with very lax rules).

rubenv:
Your concerns are 100% correct, hence my question as to whether we'd actually want to go down this path in #232.

Ended up with this due to `echo p4ss > value; kubectl create ... --from-file=value`, plain `echo` adds the new-line.

The only reason I think it makes sense is because it's accepted by `in` and `out`, not by `check`. It's the inconsistency between these three that made this so puzzling and hard to understand. I'd just as well be fine with `in` and `out` failing due to the incorrect new-line (though I have no idea how I'd write the needed fix for that).

vito:
Ah, yeah I didn't realize `in` `out` and `check` were inconsistent here! I'm guessing `in` and `out` are broken due to funky `bash`-isms.

To be honest it might be easier to just leave it as-is. It should already be consistently handled in the new/upcoming [Registry Image resource](https://github.com/concourse/registry-image-resource), which I intend to start switching everyone to over time and deprecate the Docker Image resource. The impact of this seems low enough that we could just wait for deprecation to be the solution. :)

rubenv:
Oh, right, wasn't aware that there was a replacement coming. In that case it's probably not that important.

Feel free to close!

And thanks for the awesome piece of software that Concourse is!

PR #235, opened 2018-08-21
do not start dockerd in 'in' script if skip_download is set to true

norbertbuchmueller:
The 'in' script performs no Docker operations when `skip_download` is set to true, so it does not make sense to start dockerd in that case. Considering that one usually sets `skip_download: true` to speed up the implicit "get" of the "put" operation, i think it's a meaningful optimization.


vito:
makes sense. thanks!

PR #236, opened 2018-08-22
Improve build_args description

chrishiestand:
It is inaccurate to say that build args will not persist in the final image. They are visible with `docker history` and `docker inspect`. The old description may cause someone to expose secrets.

As soon as possible, we should upgrade to docker 18.09 to support `--secret` - a way to securely pass credentials into the build context.

https://github.com/docker/cli/pull/1288


vito:
Good point, thanks!

PR #242, opened 2018-10-10
minor clarification in README regarding additional tags

glennbech:
The sentence "Path to a list" implies a file, but I managed to misunderstand it as if the parameter took a list of tags directly. I suggest this minor clarification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ version is the image's digest.
 
 #### Parameters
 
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@glennbech Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/docker-image-resource&pullRequestId=242)!

PR #251, opened 2018-12-18
Output labels as metadata

Typositoire:
I've started using labels during my docker build and found it would be pretty useful to actually see said labels as metadata.

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -94,12 +94,19 @@ fi
 echo "$repository" > ${destination}/repository
 echo "$tag" > ${destination}/tag
 echo "$digest" > ${destination}/digest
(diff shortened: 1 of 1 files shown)


vito:
@Typositoire Hey, sorry for not reviewing sooner!

I started reviewing this and then noticed this entire resource has no tests for `in`/`get`, which is pretty scary. I don't want to make you write those, though, because they should have been there already. 🙁 In any case, I'm worried about making any changes without tests and I'm hoping to phase this resource type soon in favor of the [`registry-image` resource](https://github.com/concourse/registry-image-resource) and [`builder` task](https://github.com/concourse/builder).

I'm leaning towards just closing this and beginning the process of deprecating the `docker-image` resource soon after 5.0 is out.

If you like, you could submit a PR to the `registry-image` resource instead - we just recently started adding metadata to it, so this might be easy add in and test! I promise I'll review that one faster. 🙂

Typositoire:
All good. I'll take a look at `registry-image`. I've been using my branch for a while now and it works perfectly. But I understand your point! Thanks!

PR #253, opened 2019-01-19
Small set of improvements

paroxp:
## What

The `repository` corresponding note is irrelevant when it comes to real
deal certificates. We can save some time to users who may experience
issues with their private registry not accepting the images being
pushed, due to unexpected registry string comparison.

We've made a futuristic change to the `out` script, in regards to digest
retriver.

Although the `tail -1` works in current output format, it may be a
fragile detection.

We add some functionality around signing the image with the use of
`notary`, which essentially produces 2 additional lines at the end of
the output confirming the successful signature.

It is our belief, that using a simple `grep 'digest'` will be capable of
obtaining the relevant line from the output, regardless its position.

Coincidentally, the `awk` parsing remains the same.

## How to review

- Code review should be sufficient
- Optionally, build the image and use in concourse pipeline


README.md
```diff
@@ -9,8 +9,9 @@ Note: docker registry must be [v2](https://docs.docker.com/registry/spec/api/).
 * `repository`: *Required.* The name of the repository, e.g.
 `concourse/docker-image-resource`.
 
-  Note: When configuring a private registry, you must include the port
-  (e.g. :443 or :5000) even though the docker CLI does not require it.
+  Note: When configuring a private registry **with self-signed certificate**,
```
vito: I think this is too specific - it could be using their own internal CA, not just self-signed. :slightly_smiling_face:

assets/out
```diff
@@ -284,7 +284,7 @@ if [ -e /tmp/push-failed ]; then
   exit 1
 fi
 
-digest="$(tail -1 push-output | awk '{print $3}')"
+digest="$(grep 'digest' push-output | awk '{print $3}')"
```
vito: makes sense!

README.md
paroxp: I'm trying to think of a sentence that does make sense. How about `with an "invalid" certificate`?

vito: Maybe `when configuring a private registry using a non-root CA`? A lot of organisations have their own CA for internal traffic so I wouldn't want to say it's invalid.

paroxp:
```suggestion
  Note: When configuring a private registry **using non-root CA**,
```

Sounds good!

vito:
Oh, just noticed the DCO bot isn't happy - would you mind signing your commits? Details and instructions here: https://github.com/concourse/docker-image-resource/pull/253/checks?check_run_id=52778757

paroxp:
Yeah. Done now (again*). Thanks and sorry :D

* Had to be done for an additional commit that reworded readme.

vito:
sweet, thanks and np!

PR #256, opened 2019-02-11
switch to ubuntu:bionic for base image

chenbh:
Part of concourse/concourse#3266

* resolve compatibility issue between busybox vs coreutil `timeout`
* alpine packages `docker` nicely, ubuntu requires adding docker's own repository
* Image size is now 305mb -> 448mb


assets/common.sh
```diff
@@ -93,7 +93,7 @@ start_docker() {
   declare -fx try_start
   trap stop_docker EXIT
 
-  if ! timeout -t ${STARTUP_TIMEOUT} bash -ce 'while true; do try_start && break; done'; then
```
cirocosta: There *might* be a reason why we need this `-t` 🤔 

From [`man 1 timeout`](http://man7.org/linux/man-pages/man1/timeout.1.html) it indeed says that we don't need, but it might be that busybox (what alpine uses as a replacement for these binaries) needs.

cirocosta: Oh yeah, busybox:

```sh
BusyBox v1.29.3 (2019-01-24 07:45:07 UTC) multi-call binary.

Usage: timeout [-t SECS] [-s SIG] PROG ARGS

Runs PROG. Sends SIG to it if it is not gone in SECS seconds.
Defaults: SECS: 10, SIG: TERM.
```

To get rid of that difference we can install `coreutils` (the package that `timout` lives in which `busybox` "immitates"):

```sh
apk add --update coreutils
timeout --help
Usage: timeout [OPTION] DURATION COMMAND [ARG]...
```

but that *might* be too much 🤔

chenbh: Yeah I remember seeing that `busybox` requires a separate flag for the duration whereas the `ubuntu` image just had it as a required arg.

chenbh: Oh I just realized what you were talking about, this change was originally made under the assumption that we will only be using it to build `ubuntu` images. However we have since changed that to build both `alpine` and `ubuntu` images and so this will break things.

chenbh:
pulled new version of `alpine:edge` and ...
```
$ docker run -it alpine:edge
/ # timeout --help
BusyBox v1.30.1 (2019-02-16 10:01:46 UTC) multi-call binary.

Usage: timeout [-s SIG] SECS PROG ARGS

Runs PROG. Sends SIG to it if it is not gone in SECS seconds.
Default SIG: TERM.
```

Good old busybox version bumps

cirocosta:
Hey, I still have problems here though:

```
• Failure [0.490 seconds]
Out
/go/src/github.com/concourse/docker-image-resource/tests/out_test.go:18
  when build arguments contain envvars
  /go/src/github.com/concourse/docker-image-resource/tests/out_test.go:200
    expands envvars and passes the arguments correctly to the docker daemon [It]
    /go/src/github.com/concourse/docker-image-resource/tests/out_test.go:201

    Got stuck at:
        timeout: unrecognized option: t
        BusyBox v1.30.1 (2019-02-16 10:01:46 UTC) multi-call binary.

        Usage: timeout [-s SIG] SECS PROG ARGS

        Runs PROG. Sends SIG to it if it is not gone in SECS seconds.
        Default SIG: TERM.
        Docker failed to start within 120 seconds.
        cat: can't open '/tmp/docker.pid': No such file or directory

    Waiting for:
        DOCKER ARG: --build-arg
```

chenbh:
Worked with @cirocosta and it seemed to be a flake

PR #261, opened 2019-03-14
look for registry in URLs not only in FROM, but also in ARG

### LINKED ISSUE #162: ECR pull fails when build ARG is used in FROM statement for base image
If the base image is passed in via an ARG then the regex match will not recognise that it needs to login to ECR.

e.g.
```Dockerfile
ARG BASE_IMAGE
FROM ${BASE_IMAGE:-my-other-base:latest}
...
```

```yaml
put: my-ecr
      params:
        build: source
        build_args_file: previous/build_args.json
```

```bash
Sending build context to Docker daemon  5.632kB
Step 1/3 : ARG BASE_IMAGE
 ---> 
Step 2/3 : FROM ${BASE_IMAGE:-my-other-base:latest}
Get https://<account-id>.dkr.ecr.us-west-2.amazonaws.com/v2/my-other-base/manifests/sha256:<sha256>: no basic auth credentials
```

Not sure how to resolve this one. Perhaps also take build_args into account when deciding whether to ECR login? This would also fail if the ARG had a fallback though. 

Would there be any harm in logging in if we see the ECR pattern anywhere in the Dockerfile or args file? It sucks but I can't think of a better way right now.

nzwei:
fixes #162

### DIFF
--- a/assets/out
+++ b/assets/out
@@ -308,3 +308,4 @@ jq -n "{
     { name: \"image\", value: $(echo $image_id | head -c 12 | jq -R .) }
   ]
 }" >&3
+
--- a/assets/out
+++ b/assets/out
@@ -308,4 +308,3 @@ jq -n "{
     { name: \"image\", value: $(echo $image_id | head -c 12 | jq -R .) }
   ]
 }" >&3
-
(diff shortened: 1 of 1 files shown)


nzwei:
fixes #162

PR #265, opened 2019-04-01
Multi target repo for build task

wabula:
Replicating behaviour of `tag_file` allow the `put` target repository to be driven from concourse task.

Image tag is generated within a concourse job, passed to Put as output.
PR allows for the repository to be defined also. 
Example, if multi repos are to be used depending on if final built image is to be label as a PR or a Release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,8 +262,7 @@ version is the image's digest.
 * `target_name`: *Optional.*  Specify the name of the target build stage.
   Only supported for multi-stage Docker builds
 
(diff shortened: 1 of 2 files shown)


bhutkovskyysos:
Any updates for this? Currently, I need to specify multiple duplicate resources just so I can push image to different repos. Repo file is a great solution to this.

vito:
I don't think we should support this, at least not right now. This resource type is in the process of being phased out. This also makes the narrative of this resource kind of weird, and just turns it into a 100% configurable blob that doesn't concretely map to anything in the real world (as resources are intended to).

PR #268, opened 2019-05-06
bug: issue 267 alpine docker build fails

### LINKED ISSUE #267: bug: alpine docker build fails
At commit 502cbb9c2578a1f6e0d3b5c131f99f1b2b49c949, `docker build -t docker-image-resource -f dockerfiles/alpine/Dockerfile .` fails while the equivalent for Ubuntu succeeds.

Here is the snippet showing the failure mode:
```
Step 10/17 : RUN apk --no-cache add       bash       docker       jq       ca-certificates       xz       util-linux     ;
 ---> Running in 2cef622f411f
fetch http://dl-cdn.alpinelinux.org/alpine/edge/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/edge/community/x86_64/APKINDEX.tar.gz
(1/27) Installing ncurses-terminfo-base (6.1_p20190420-r0)
<snip>
(6/27) Installing libcrypto1.1 (1.1.1b-r1)
ERROR: libcrypto1.1-1.1.1b-r1: trying to overwrite etc/ssl/openssl.cnf owned by libressl2.6-libcrypto-2.6.3-r0.
<snip>
(27/27) Installing xz (5.2.4-r0)
Executing busybox-1.27.2-r7.trigger
Executing ca-certificates-20190108-r0.trigger
1 error; 292 MiB in 38 packages
The command '/bin/sh -c apk --no-cache add       bash       docker       jq       ca-certificates       xz       util-linux     ;' returned a non-zero code: 1
```

Using alpine 3.9 allows the build to progress farther, however the unit tests then fail.

hstenzel:
There should be no differences in the `builder` stage between the ubuntu and
alpine docker images. The only difference should be in the resource stage
which encapsulates all of the distro specific differences.

Fixes #267

### DIFF
--- a/dockerfiles/alpine/Dockerfile
+++ b/dockerfiles/alpine/Dockerfile
@@ -1,5 +1,5 @@
 # stage: builder
-FROM golang:alpine AS builder
+FROM concourse/golang-builder AS builder
 
 COPY . /go/src/github.com/concourse/docker-image-resource
 ENV CGO_ENABLED 0


vito:
:thinking: I'm not sure why ubuntu and alpine use different builder images, but it at least doesn't seem related to #267 - that failure happens in the later builder stage, and building also seems to work fine for me locally. Maybe it was a flake?

PR #271, opened 2019-06-07
Add new param: tag_command

vlandemaine-orange:
Adding new parameter : tag_command
Setting a tag with a command (like value in file etc..)
Example : 

Get the value of "head_name" in the file metadata.json without quotes
tag_command: $(jq '.[] | select(.name=="head_name").value' pull-request/.git/resource/metadata.json | sed 's/\"//g')


defo89:
Hi @vlandemaine-orange any chance that you will merge this PR?

I think it's pretty useful and would help in use cases like:

```
- task: do something
  config:
     platform: 'linux'
     image_resource:
       type: docker-image
       source:
         repository: terraform
         tag_command: $(cat terraform.image.version/version)
```

Thanks!

PR #272, opened 2019-06-26
fix: dropping versions in check

cappyzawa:
I fixed dropping versions in check. (based on https://github.com/concourse/concourse/issues/3910#issuecomment-495771128)

I know that this Resource is deprecated and `registry-image-resource` is recommended, but I still want to use this resource because our internal Docker Registry does not support oci.

I confirmed that this change solved the problem.


cappyzawa:
Sorry, I misunderstood.

PR #290, opened 2020-02-18
output metadata that exists even when skip_download is true

### LINKED ISSUE #289: metadata not show when skip_download=true
with skip_download=true, docker is not started (which is nice), but also the variable `image_id` is not set (that part is skipped).

but later
https://github.com/concourse/docker-image-resource/blob/master/assets/in#L98-L105
the output metadata is based on image_id ... and is shown empty 
image

How about setting the metadata based on the variables that it certainly knows about even with skip_download=true?

gdamjan:
output the repository, digest and tag as metadata too

fixes #289

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -100,6 +100,8 @@ jq -n "{
     digest: $(echo $digest | jq -R .)
   },
   metadata: [
+    { name: \"repository\", value: $(echo $repository | jq -R .) },
+    { name: \"tag\", value: $(echo $tag | jq -R .) },
     { name: \"image\", value: $(echo $image_id | head -c 12 | jq -R .) }
   ]
-}" >&3
+}" | jq '{version: .version} + {metadata: [.metadata[] | select(.value != "")]}' >&3


xtremerui:
Hi @gdamjan , thank you for this PR and sry for late reply. 

If you are still intereted at getting this PR through, could you please fix the [test failure](https://ci.concourse-ci.org/builds/927445#L5e84d33d:1270:1280)?

gdamjan:
> Hi @gdamjan , thank you for this PR and sry for late reply.
> 
> If you are still intereted at getting this PR through, could you please fix the [test failure](https://ci.concourse-ci.org/builds/927445#L5e84d33d:1270:1280)?

I'm confused about that. it seems to fail on master too.

changing the tag from  "aklsdf123" to "123456789" makes the test pass on master. Should I add that to this PR too?

gdamjan:
```
Tag                json.Number     `json:"tag"`
```
was introduced in 411fdc22 3+ years ago, I can only assume the behaviour of encoding/json changed in the meanwhile??

xtremerui:
https://github.com/golang/go/issues/34472 has the context about the failure we have in master now. TLDR go behaviour changed for `json.Number`.

master should be fixed before your PR got merged. I am working on it.

xtremerui:
@gdamjan I haved fixed the master so in this PR you shouldn't need to change the test anymore. Sry I should've checked master first.

gdamjan:
I guess it's alright now.

assets/in
```diff
@@ -100,6 +100,9 @@ jq -n "{
     digest: $(echo $digest | jq -R .)
   },
   metadata: [
+    { name: \"repository\", value: \"${repository}\" },
+    { name: \"digest\", value: \"${digest}\" },
+    { name: \"tag\", value: \"${tag}\" },
```
xtremerui, line 105:
```suggestion
    { name: \"repository\", value: $(echo $repository | jq -R .) },
    { name: \"tag\", value: $(echo $tag | jq -R .) },
```
`digest` is already showing in the version so I think we don't need to include it in metadata again.

Is there an easy way to omit `image` when its value is null if `skip_download` is true?

gdamjan: > digest is already showing in the version so I think we don't need to include it in metadata again.

if it can be used in the next step easily, then ok.

xtremerui: Another reason behind is concourse/registry-image-resource always show `repository` and `tag` both in `in` and `out` without `image_id` and `digest` regardlessly. Since these two resource types are basically serving the same purpose I feel unifying the interface would help user to migrate from docker-image-resource to registry-image-resource.

gdamjan:
Empty values from the metadata can be filtered by piping the metadata output to:
```
| jq '{version: .version} + {metadata: [.metadata[] | select(.value != "")]}'
```

xtremerui:
Thank you for the PR!

PR #300, opened 2020-09-16
use HEAD request to avoid rate limits

vito:
fixes https://github.com/concourse/concourse/issues/6074

### DIFF
--- a/cmd/check/main.go
+++ b/cmd/check/main.go
@@ -5,7 +5,6 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


cmd/check/main.go
```diff
@@ -20,12 +19,11 @@ import (
 	ecr "github.com/awslabs/amazon-ecr-credential-helper/ecr-login"
 	ecrapi "github.com/awslabs/amazon-ecr-credential-helper/ecr-login/api"
 	"github.com/concourse/retryhttp"
-	"github.com/docker/distribution"
 	"github.com/docker/distribution/digest"
 	_ "github.com/docker/distribution/manifest/schema1"
 	_ "github.com/docker/distribution/manifest/schema2"
 	"github.com/docker/distribution/reference"
-	"github.com/docker/distribution/registry/api/v2"
+	v2 "github.com/docker/distribution/registry/api/v2"
```
xtremerui, line 26: Is this change needed?

vito: It was done automatically by `goimports` so I assume it's fine. Maybe so it doesn't get confused with go module versioning conventions which ignore the v2 suffix - in this case it's just a regular package name so it must have decided to explicitly name it that way.

xtremerui (approved):

PR #302, opened 2020-09-23
[feature] add  configuration

brainpoint:
Add `source.official_registry` configuration

Allow users to change the default official_registry url

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@ Note: docker registry must be [v2](https://docs.docker.com/registry/spec/api/).
 
 * `registry_mirror`: *Optional.* A URL pointing to a docker registry mirror service.
 
+* `official_registry`: *Optional.* The default offcial registry is registry-1.docker.io
+
 * `ca_certs`: *Optional.* An array of objects with the following format:
 
   ```yaml
--- a/cmd/check/main.go
+++ b/cmd/check/main.go
@@ -30,6 +30,8 @@ import (
 	"github.com/pivotal-golang/clock"
 )
 
+var officialRegistry = "registry-1.docker.io"
+
 func main() {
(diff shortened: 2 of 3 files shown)


vito:
Thanks for the PR! I don't think it we'll merge this though because the `docker-image` resource type is deprecated - we're trying to limit changes to bug fixes and maintenance. You may want to try out the [`registry-image` resource type](https://github.com/concourse/registry-image-resource/) which will be its eventual replacement.

I'm curious, what is the use of setting this rather than putting the registry in the `repository` field?

i.e. rather than...:

```yaml
repository: foo
official_registry: my-registry.com
```

... it seems like you could set:

```yaml
repository: my-registry.com/foo
```

brainpoint:
> Thanks for the PR! I don't think it we'll merge this though because the `docker-image` resource type is deprecated - we're trying to limit changes to bug fixes and maintenance. You may want to try out the [`registry-image` resource type](https://github.com/concourse/registry-image-resource/) which will be its eventual replacement.
> 
> I'm curious, what is the use of setting this rather than putting the registry in the `repository` field?
> 
> i.e. rather than...:
> 
> ```yaml
> repository: foo
> official_registry: my-registry.com
> ```
> 
> ... it seems like you could set:
> 
> ```yaml
> repository: my-registry.com/foo
> ```

Thanks,

The parameter `official_registry` can be used in areas with a great firewall, such as China; We can keep concourse-ci scripts as far as possible consistent by only to set `official_registry`.

vito:
@brainpoint Hmm, I think `registry_mirror` would be more appropriate for that use case. With https://github.com/concourse/concourse/pull/6079 (coming in v6.7) you'll even be able to set it as a default for everything without having to update all your pipelines.

Issue #305, opened 2020-10-15
'registry_mirror' indiscriminately overwrites a 'repository' containing a registry hostname


mdb:
In current implementation, if a `registry_mirror` is configured, images whose `source.repository` contains a registry-host-prefixed image (such as `my-registry.com/foo/bar`) will still be checked at the `registry_mirror` registry and _not_ the registry specified in the `repository` value. This may cause the `check` to fail if the configured `registry_mirror` does not mirror the registry host specified in the `repository` value, but instead mirrors a _different_ registry.

This is especially problematic when a Concourse-wide [default registry has been configured](https://github.com/concourse/concourse/pull/6079). 

For example, given the following default `docker-image` configuration:

```yaml
docker-image:
  registry_mirror: https://my-default-mirror.com
```

...and the following `docker-image-resource`:

```yaml
- name: my-image
  type: docker-image
  source:
    repository: my-registry.com/foo/bar
```

...the image is checked at `my-default-mirror.com/foo/bar`. Unless `my-default-mirror.com` is a mirror to `my-registry.com`, the `check` invocation may result in failures like the following:

```
run check step: run check step: check: resource script '/opt/resource/check []' failed: exit status 1

stderr:
failed to fetch manifest: Head "https://my-default-mirror.com/v2/foo/bar/manifests/latest": unauthorized: BAD_CREDENTIAL
```

This prevents individual users from using `docker-image-resource` images sourced from any registry other than the registry mirrored by the default `registry_mirror`. This is particularly problematic in large Concourse installations -- such as the one my team operates, which services thousands of users -- when individual pipelines, jobs, and tasks may use `docker-image-resource` instances sourced from their own registries, or even public registries that are not mirrored by the default `registry_mirror`.

It's worth noting that this is _not_ true of `docker`, nor the [registry-image-resource](https://github.com/concourse/registry-image-resource), both of which _can_ successfully resolve `my-registry.com/foo/bar` from `my-registry.com`, even when `my-default-mirror.com` has been configured as a registry mirror. In the case of `docker`, it appears `my-registry.com/foo/bar` is successfully resolved without performing any requests to `my-default-mirror.com`.

vito:
fixed with #306 - thanks!

PR #306, opened 2020-10-15
Prevent 'registry_mirror' from overwriting the hostname in a 'repository' containing a registry host name

mdb:
This addresses issue #305 

I'm also happy to ~write tests~ and/or provider further tweaks, but thought it might be worth getting some early feedback before doing so. Thanks!

Update: I went ahead and wrote a few tests :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ Note: docker registry must be [v2](https://docs.docker.com/registry/spec/api/).
 * `repository`: *Required.* The name of the repository, e.g.
 `concourse/docker-image-resource`.
 
-  Note: When configuring a private registry which requires a login the 
-  registry's hostname must contain at least one '.' e.g. `registry.local`. 
+  Note: When configuring a private registry which requires a login the
+  registry's hostname must contain at least one '.' e.g. `registry.local`.
   Otherwise docker hub will be used.
   Note: When configuring a private registry **using a non-root CA**,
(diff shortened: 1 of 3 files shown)


mdb:
@vito Apologies; I posted a [similar question](https://github.com/concourse/registry-image-resource/pull/248#issuecomment-712275928) pertaining to the `registry-image-resource`. I'm curious to hear your thoughts on this and whether, if approved, a new `docker-image-resource:latest` might be released within a comfortable window ahead of the [November 1, 2020 rate limits](https://www.docker.com/blog/scaling-docker-to-serve-millions-more-developers-network-egress/) Docker Hub plans to impose.

I ask because I believe the ability to configure a default `docker-image-resource.registry_mirror` will help mitigate the impending [November 1, 2020 rate limiting](https://www.docker.com/blog/scaling-docker-to-serve-millions-more-developers-network-egress/). However, I'm concerned that doing so without the changes proposed in this PR will have undesired consequence for those users of the `docker-image-resource` whose configurations include a `repository` containing explicitly declared registry-host-prefixed image (such as `my-registry.com/foo/bar`).

vito (approved):
Thanks! I'll go ahead and merge this too.

For anyone following along at home, there was a hefty conversation in https://github.com/concourse/registry-image-resource/pull/248 which ultimately concluded with "yep makes sense."

PR #308, opened 2020-10-19
check: use GET if if HEAD doesnt return digest

### LINKED ISSUE #307: PR 300 causes issues with "redhat" repository
The recent PR 300 which uses HEAD to check the resources causes the following errors when access "redhat" image repository.

"run check step: run check step: check: resource script '/opt/resource/check []' failed: exit status 1

stderr:
no digest header returned"

We have the resource configured like this ...

```
  - name: base-image-logdna-agent
    source:
      repository: registry.access.redhat.com/ubi8/nodejs-12
      tag: latest
    type: docker-image
```

We had to revert to using 1.3.4 of the docker-image-resource for all "redhat" repositories.  We need this PR to be rolled back or fixed so that this is not the "default" and that we can perhaps have a source property to define that we do NOT use HEAD when doing the check for certain repositories.

vito:
fixes #307

This is necessary for registries that don't return a `Docker-Content-Digest` header. I had hoped that they'd all be conforming by now. No such luck.

### DIFF
--- a/cmd/check/main.go
+++ b/cmd/check/main.go
@@ -5,6 +5,7 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


YoussB (approved):
LGTM! The code looks good, and I tested the image against the configuration in the issue and it is working 🚀

PR #312, opened 2020-10-30
Fix Check implementation to properly use HEAD requests

### LINKED ISSUE #311: As of v1.5.0 only GET requests are made not HEAD
I looks like in v1.5.0 when the HEAD requests were broken out into their own function the logic want updated to properly call them. This results in the old behavior of using GET requests for everything which happened to break one of our workflows. The issue I believe can be traced to [here](https://github.com/concourse/docker-image-resource/blob/5141daf2352d402e14d75cfeb2da3286d8c4640a/cmd/check/main.go#L94-L108) where `headDigest` needs to be called instead of `fetchDigest` such that if headDigest fails it cascades into calling fetchDigest.

```golang
	latestDigest, foundLatest := fetchDigest(client, latestManifestURL, request.Source.Repository, tag)


	if request.Version.Digest != "" {
		digestRef, err := reference.WithDigest(namedRef, digest.Digest(request.Version.Digest))
		fatalIf("failed to build cursor manifest URL", err)


		cursorManifestURL, err := ub.BuildManifestURL(digestRef)
		fatalIf("failed to build manifest URL", err)


		cursorDigest, foundCursor := fetchDigest(client, cursorManifestURL, request.Source.Repository, tag)


		if foundCursor && cursorDigest != latestDigest {
			response = append(response, Version{cursorDigest})
		}
	}
```

daviddob:
Fixes #311

This changes the check flow to call headDigest instead of fetchDigest directly.
If headDigest fails to get the digest it will fall back to using fetchDigest.

Tested in the following scenarios:
- Upstream to dockerhub with library and user repos
- Against private nexus registry with pull-through cache both directly and with registry_mirror
- Against the repo called out in #307 which spawned the initial change.

This fix is required for us to upgrade to 6.7.x as the nexus pull-through-cache works with HEAD requests and not GET requests for check
(problem for another day) as well as important for the Dockerhub ratelimiting on Nov 2

### DIFF
--- a/cmd/check/main.go
+++ b/cmd/check/main.go
@@ -91,7 +91,7 @@ func main() {
 	latestManifestURL, err := ub.BuildManifestURL(taggedRef)
 	fatalIf("failed to build latest manifest URL", err)
 
-	latestDigest, foundLatest := fetchDigest(client, latestManifestURL, request.Source.Repository, tag)
+	latestDigest, foundLatest := headDigest(client, latestManifestURL, request.Source.Repository, tag)
 
 	if request.Version.Digest != "" {
 		digestRef, err := reference.WithDigest(namedRef, digest.Digest(request.Version.Digest))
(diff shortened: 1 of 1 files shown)


vito:
thanks!

PR #314, opened 2020-11-08
Skip Systemd cgroup creation on systems with OpenRC

glorpen:
In short, forcefully creating systemd cgroups on systems without systemd makes Docker fail with same error as in https://github.com/concourse/docker-image-resource/commit/264b3718b5b85d850c208159fe2dd364620a83d1  - `docker: Error response from daemon: cgroups: cannot find cgroup mount destination: unknown.` just this time it hapens on host system.
( Previously reported as https://github.com/vito/oci-build-task/issues/27 )

To make things more fun, some systems have cgroups auto-cleaning, so after container is destroyed everything will work again.

This PR tries to fix this behavior on OpenRC systems, my knowledge about cgroups is shallow but it seems to not break anything.

How to reproduce problem:

Vagrantfile:
```
$script = <<-SCRIPT
apk add docker
/etc/init.d/docker start
SCRIPT

Vagrant.configure("2") do |config|
  config.vm.define "alpine" do |config|
  config.vm.hostname = "test"
  config.vm.box = "generic/alpine312"
  config.vm.box_check_update = false
  config.vm.provider :libvirt do |v|
    v.memory = 1024
    v.cpus = 1
    end
  end
  config.vm.provision "shell", inline: $script
end
```

In first terminal:
```
test:/home/vagrant# docker run --rm -it alpine:3.11 echo test
test
```
In second terminal:
```
docker run --rm -it --privileged concourse/docker-image-resource bash -c "echo '"'{"source":{"repository":"alpine","tag":"3.11"}}'"' | /opt/resource/in /example; sleep 30s"
```
And then back in first terminal:
```
test:/home/vagrant# docker run --rm -it alpine:3.11 echo test
docker: Error response from daemon: cgroups: cannot find cgroup mount destination: unknown.
ERRO[0001] error waiting for container: context canceled
```

After 30s, container with `docker-image-resource` will exit, systemd cgroup will be removed and hosts docker will work again.

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -41,7 +41,7 @@ sanitize_cgroups() {
     fi
   done
 
-  if ! test -e /sys/fs/cgroup/systemd ; then
+  if [ ! -e /sys/fs/cgroup/systemd ] && [ $(cat /proc/self/cgroup | grep '^1:name=openrc:' | wc -l) -eq 0 ]; then
     mkdir /sys/fs/cgroup/systemd
     mount -t cgroup -o none,name=systemd none /sys/fs/cgroup/systemd
   fi


muntac:
@glorpen I tried reproducing the issue in a pipeline using `gentoo/stage3` and `alpine:3.11` but did not see the error. When I go into those containers I see they're using `systemd`.

Are your workers running Gentoo as well, and does that affect the outcome? My workers were a local docker-compose one, and a GKE deployed one.

glorpen:
@muntac for this problem to happen workers have to run on non-systemd VMs, from the way you written `alpine:3.11` I'm guessing that it was a container :). As containers don't usually have real init system, it does not matter what container it is - code creates cgroup (in this case `systemd`) that does not exist on host so docker errors out.

Example lite:
```
test:/home/vagrant# cat /proc/1/cgroup | grep systemd
test:/home/vagrant# docker run --rm -it --privileged alpine:3.11 sh -c 'mkdir /sys/fs/cgroup/systemd; mount -t cgroup -o none,name=systemd none /sys/fs/cgroup/systemd; sleep 30s' &
[2] 5994
test:/home/vagrant# docker run --rm -it alpine:3.11 echo test
docker: Error response from daemon: cgroups: cannot find cgroup mount destination: unknown.
test:/home/vagrant# kill -9 %1
test:/home/vagrant# docker run --rm -it alpine:3.11 echo test
test
```
In above example, you could change `systemd` to any other string that is unknown to host and would get same result.

muntac (approved):

muntac:
@glorpen Makes sense. Was able to reproduce. I had to add the version and digest fields to the `in` script as I guess those come from the `check`.

Did acceptance testing on both alpine and ubuntu vms, and everything seems good. Thanks for the PR!

PR #315, opened 2020-11-12
Harry/ch87243/make our concourse docker image resource

harry-plutoflume:


harry-plutoflume:
I committed this from our fork to upstream by mistake - sorry!

Issue #278, opened 2019-08-22
Provide credentials for base image different from the image to be built


chbiel:
Hi,
i have the following setup:

Dockerfile of "child" image
```
FROM quay.io/team/base
...do something...
```

quay.io need authentication and the following would work as expected.
```
- name: docker-image-tools
    type: docker-image
    icon: docker
    source:
      repository: quay.io/team/base
      username: ((quay_username))
      password: ((quay_password))
```

Now I want to build the "child" image above, but push it to AWS ECR:
```
- name: docker-image-child
    type: docker-image
    icon: docker
    source:
      repository: account.dkr.ecr.region.amazonaws.com/child
```

When I execute that I get an error that authentication for quay.io is missing (what is basically OK).

But now the question:
Can I provide different credentials for the base image and the image to created/push? Or do i have to ensure that the base image and the target image are in the same repo?

mischcon:
I have the same issue and I am wondering why this issue does not affect many more people?


Would it somehow be possible to provide the content of the '.docker/config.json' file to the resource? This way one could craft such a config.json file that contains all needed credentials.

andrea3bianchi:
This makes this resource completely unusable when pushing to a different repository now that dockerhub pretty much requires authentication due to rate limiting as you can't authenticate to dockerhub to build the image and push to any different repository

Issue #243, opened 2018-10-10
WARNING! Your password will be stored unencrypted in /root/.docker/config.json


chrishiestand:
Using recent versions of the resource, e.g. 

```
resource_types:

- name: docker-image
  type: docker-image
  privileged: true
  source:
    repository: concourse/docker-image-resource
    tag: pr-208
```

Which contains docker version:
```
bash-4.4# docker version
Client:
 Version:           18.06.1-ce
 API version:       1.38
 Go version:        go1.11
 Git commit:        3b19bc6ba8
 Built:             Mon Sep  3 10:21:47 2018
 OS/Arch:           linux/amd64
 Experimental:      false
```

Using basic authentication we now get a docker warning in the build step:

> WARNING! Your password will be stored unencrypted in /root/.docker/config.json.
> Configure a credential helper to remove this warning. See
> https://docs.docker.com/engine/reference/commandline/login/#credentials-store

I suppose the actual security hasn't changed at all, but the warning will freak people out. This may be a compelling reason to abandon the local docker daemon approach sooner rather than later?

waswrongassembled:
Yes, the warning freaks people out! :(

vito:
> This may be a compelling reason to abandon the local docker daemon approach sooner rather than later?

Yep! I'm gonna close this out since we're already phasing out `docker-image` resource in favor of the [`registry-image` resource](https://github.com/concourse/registry-image-resource) which doesn't use a Docker daemon and doesn't have this problem.

The `docker-image` resource will live on for a while longer until concourse/rfcs#7 is done and a task can be used just as nicely instead, so until then we'll just live with these warnings for image building.

PR #317, opened 2020-11-24
Added docker_config_json paraneter

markround:
This means you can specify a full ~/.docker/config.json file to use for
authentication, which is useful to work around Docker Hub's rate-limiting.

You can pass this in (e.g. from a secrets source such as Credhub or k8s
secrets), and authenticate against multiple registries.

Also added GODEBUG="x509ignoreCN=0" env var to work around "legacy CN"
format SSL certificates warning, introduced with latest version of Go SSL
library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ Note: docker registry must be [v2](https://docs.docker.com/registry/spec/api/).
 * `aws_access_key_id`: *Optional.* AWS access key to use for acquiring ECR
   credentials.
 
+* `docker_config_json` : *Optional.* The raw `config.json` file used for authenticating with Docker registries. If specified, `username` and `password` parameters will be ignored. You may find this useful if you need to be authenticated against multiple registries (e.g. pushing to a private registry, but you also also need to pull authenticate to pull images from Docker Hub without being rate-limited).
+
 * `aws_secret_access_key`: *Optional.* AWS secret key to use for acquiring ECR
   credentials.
 
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -2,6 +2,9 @@ LOG_FILE=${LOG_FILE:-/tmp/docker.log}
 SKIP_PRIVILEGED=${SKIP_PRIVILEGED:-false}
 STARTUP_TIMEOUT=${STARTUP_TIMEOUT:-120}
 
+# Otherwise we get "certificate relies on legacy Common Name field"
+export GODEBUG="x509ignoreCN=0"
+
 sanitize_cgroups() {
   mkdir -p /sys/fs/cgroup
   mountpoint -q /sys/fs/cgroup || \
@@ -205,4 +208,10 @@ docker_pull() {
 
   printf "\n${RED}Failed to pull image %s.${NC}" "$1"
   return 1
+}
+
+docker_config_json_to_file() {
+  local docker_config_json="${1}"
+  mkdir -p ~/.docker
+  echo "${1}" > ~/.docker/config.json
 }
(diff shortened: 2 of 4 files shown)


markround:
This is a work-around for #278 which has recently bitten me due to Docker Hub's rate-limiting. I needed to be able to drop a raw docker `config.json` in so I can authenticate against multiple registries e.g. log into Docker Hub so I don't get rate-limited when pulling base images, but also authenticate against my target private registry.

Hope this helps someone in any case!

xtremerui:
Hi, thx for the PR. Have you tried https://github.com/concourse/registry-image-resource? We encourage users to move to `registry-image-resource` rather than adding new features to this resource type (though this PR falls between new feature and bug fix as the change is introduced from docker itself)

If that works in your case I'd rather hold on this PR since all your credetials will be stored plain text in the config file, refer to https://github.com/concourse/docker-image-resource/issues/243

markround:
@xtremerui The registry-image-resource doesn't really work for me, as it doesn't deal with building containers using public base images. Which is where I was being badly bitten by Docker Hub's rate-limiting when not authenticating as I'm behind CGNAT and presumably sharing my public IP with 100s/1000s of other users. 

I could I guess hook registry-image up to the oci-build resource but that's somewhat in flux at the moment and doesn't really offer anything at the moment over this resource as far as I can see (e.g. it still requires `privileged: true`, although I think that's planned to change in the future). 

I understand the hesitation to merge given the config file being written to disk, but then as far as I understand it, that's still what happens with the "docker login" approach. Although I agree it's not best practice to leave credentials floating around on disk. 

I'm fine with whatever you choose to do though :) 

The fact that this PR exists and is linked in the associated issue is enough for me, hopefully it'll help someone else with a workaround in the worst-case scenario.

README.md
```diff
@@ -27,6 +27,8 @@ Note: docker registry must be [v2](https://docs.docker.com/registry/spec/api/).
 * `aws_access_key_id`: *Optional.* AWS access key to use for acquiring ECR
   credentials.
 
+* `docker_config_json` : *Optional.* The raw `config.json` file used for authenticating with Docker registries. If specified, `username` and `password` parameters will be ignored. You may find this useful if you need to be authenticated against multiple registries (e.g. pushing to a private registry, but you also also need to pull authenticate to pull images from Docker Hub without being rate-limited).
```
xtremerui, line 30:
```suggestion
* `docker_config_json` : *Optional.* The raw `config.json` file used for authenticating with Docker registries. If specified, `username` and `password` parameters will be ignored. You may find this useful if you need to be authenticated against multiple registries (e.g. pushing to a private registry, but you also need to pull authenticate to pull images from Docker Hub without being rate-limited).
```

assets/out
```diff
@@ -52,7 +53,12 @@ start_docker \
 	"${max_concurrent_uploads}" \
 	"$insecure_registries" \
 	"$registry_mirror"
-log_in "$username" "$password" "$registry"
+  
+if [ -z "$docker_config_json" ]; then
+  log_in "$username" "$password" "$registry"
+else
+  docker_config_json_to_file "$docker_config_json"
+fi
```
xtremerui, line 61: Could you add an out test to cover this feature?

assets/common.sh
```diff
@@ -2,6 +2,9 @@ LOG_FILE=${LOG_FILE:-/tmp/docker.log}
 SKIP_PRIVILEGED=${SKIP_PRIVILEGED:-false}
 STARTUP_TIMEOUT=${STARTUP_TIMEOUT:-120}
 
+# Otherwise we get "certificate relies on legacy Common Name field"
+export GODEBUG="x509ignoreCN=0"
```
xtremerui, line 6: In what user case that you need this env var? Are you seeing the error when check/in/out?

From what I understand in this resource, only the check cmd is using Go, so in/out that uses bash script should not be affected?

PR #321, opened 2021-01-18
Feature: build image once

lootbox:
This feature is for that cases if you don't want to build image it it already exists in the image registry.
By default it not changes behavior. To enable this feature was added a new flag `build_once`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,8 @@ version is the image's digest.
   { "EMAIL": "me@yopmail.com", "HOW_MANY_THINGS": 1, "DO_THING": false }
   ```
 
+* `build_once`: *Optional.* If this parameter is set to "true", the resource will check the image existence in the registry to avoid rebuilding unnecessarily the image.
+  
 * `cache`: *Optional.* Default `false`. When the `build` parameter is set,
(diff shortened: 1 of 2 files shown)


taylorsilva:
Hey @lootbox, thanks for the PR! 

I wanted to let you know that we plan to deprecate this resource in the future. It's being replaced by the [registry image resource](https://github.com/concourse/registry-image-resource/) (for fetching and pushing images) and the [oci-build task](https://github.com/vito/oci-build-task) (for building images). You can see how these two objects work together in this blog post: https://blog.concourse-ci.org/how-to-build-and-publish-a-container-image/

Both the registry-image resource and the oci-build task are also more performant than the docker resource because neither of them need to spin up docker in order to perform their container-related functions. They're docker-less!

Since we plan to deprecate this resource we've been very cautious to add new features as they increase our support burden. I'm wondering if there's any reason you're still using this resource instead of the registry image + oci-build task? Also, is this build once feature in your PR something that would be helpful for you in the registry-image resource?

taylorsilva:
Closing because there's been no update.

PR #323, opened 2021-01-22
Output contents of log file on error

bartlettt:
This PR outputs the contents of the log file to console in the event docker fails to start. A lot of my users are having issues with docker starting, and troubleshooting is difficult without this information. The containers have often been garbage collected by the time I come to investigate.

Signed-off-by: Tom Bartlett <tom.bartlett@spire.com>

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -94,7 +94,7 @@ start_docker() {
   trap stop_docker EXIT
 
   if ! timeout ${STARTUP_TIMEOUT} bash -ce 'while true; do try_start && break; done'; then
-    cat "${LOG_FILE}"
+    [ -f "$LOG_FILE" ] && cat "${LOG_FILE}"
     echo Docker failed to start within ${STARTUP_TIMEOUT} seconds.
     return 1
   fi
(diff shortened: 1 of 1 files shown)


vito (approved):
makes sense - thanks!

vito:
Looks like this causes the tests to hang because they set `LOG_FILE` to `/dev/stderr`, and `cat /dev/stderr` will get stuck. I'm putting together a PR to set it to a tempfile instead.

assets/common.sh
```diff
@@ -94,6 +94,7 @@ start_docker() {
   trap stop_docker EXIT
 
   if ! timeout ${STARTUP_TIMEOUT} bash -ce 'while true; do try_start && break; done'; then
+    cat "${LOG_FILE}"
```
vito, line 97:
```suggestion
    [ -f "$LOG_FILE" ] && cat "${LOG_FILE}"
```

vito (changes requested):
Actually, that fails because the tests actually test against the output written to `/dev/stderr`.

Maybe this could just defend against catting a non-file instead?

vito:
Thanks!

PR #331, opened 2021-03-30
Add support for securely passing ssh identity to docker build

ericb-summit:
Some Dockerfile builds require ssh authentication to perform some operations.  Most notably pulling dependencies from private repos.  To avoid leaking the ssh credentials in the build layers, docker buildkit supports the `--ssh default` argument which does a pass-thru of the ssh-agent socket from the host to the builder.  This PR defines a new parameter `ssh_identity` and if set, it temporarily spawns an `ssh-agent` during the docker build.


xtremerui:
Hi could you add a out test in tests/out_test.go if possible?

JasonDictos:
This fix has been re-implemented here: https://github.com/concourse/docker-image-resource/pull/364

PR #332, opened 2021-03-30
Fix broken interaction of BUILDKIT and docker ecr_login plugin

ericb-summit:
When running within docker-in-docker context, and more specifically in this resource, and only when buildkit is enabled, there is an issue with the aws `ecr_login` authenticator plugin of docker.  Specifically, it gets confused and tries to authenticate attempts to fetch from non-ECR registries.  It then gives up errors such as "xxx... is not a valid repository URI for Amazon Elastic Container Registry" for layers from say, docker hub, and the `docker build` fails.

Fortunately, there is already code in `assets/out:242` which does an explicit pull for base images from ECR, if any, which means, the `docker build` step doesn't actually need the `ecr_login` plugin.

This PR removes the `ecr_login` plugin while calling `docker build`, and then puts it back immediately after.

I have tested that pushing to ECR still works after this patch, so the plugin still works.

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -118,7 +118,9 @@ log_in() {
     echo "${password}" | docker login -u "${username}" --password-stdin ${registry}
   else
     mkdir -p ~/.docker
-    echo '{"credsStore":"ecr-login"}' >> ~/.docker/config.json
+    touch ~/.docker/config.json
+    # This ensures the resulting JSON object remains syntactically valid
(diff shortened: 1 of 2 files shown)


taylorsilva (approved):
I'm not able to test because I don't have access to an AWS infra. LGTM though as a minor fix and seems it won't break existing workflows if I understand your comment correctly.

PR #333, opened 2021-04-17
Expand build args in FROM computation

xguerin:
Expands build args used in `FROM` statements before the ECR pattern match.

It enables the following to work with ECR repositories:
```dockerfile
ARG REGISTRY
ARG NAMESPACE
FROM ${REGISTRY}/${NAMESPACE}/some-image
```

It uses `echo -E` to handle cases where build arg values contain escape sequences.

### DIFF
--- a/assets/out
+++ b/assets/out
@@ -172,9 +172,9 @@ elif [ -n "$build" ]; then
     fi
   done
 
-  build_arg_keys=($(echo "$build_args" | jq -r 'keys | join(" ")'))
-  if [ "${#build_arg_keys[@]}" -gt 0 ]; then
-    for key in "${build_arg_keys[@]}"; do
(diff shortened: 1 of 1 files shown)


EstebanFS:
Hey thanks for the PR, is there any reason why you are not using the [OCI build task](https://github.com/vito/oci-build-task) + the [registry image resource](https://github.com/concourse/registry-image-resource)? We want to understand your use-case because we want to deprecate this resource eventually. Is this something that could be added to one or both of those resource/task images?

xguerin:
Hi, thanks for the pointers. I was not aware of these resources. They may very well be addressing my use-case. I'll check them out and get back to you.

taylorsilva:
@xguerin Hey, wondering if the OCI build task + registry image were able to meet your requirements?

taylorsilva:
Closing because there's been no update.

PR #336, opened 2021-05-27
Skip cgroup sanitization for cgroup v2

rroohhh:
On cgroup v2 systems (like fedora v31+) the cgroup sanitization will fail with a error message like
```
mount: /sys/fs/cgroup/cpuset: cgroup already mounted on /sys/fs/cgroup.
```

(The exact mountpoint probably depends on the available cgroups).

This simply bails out, as this mounting is no longer necessary with cgroup v2.

This was tested on a cgroup v2 system using a containerd runner.


vito:
This resource type is deprecated but not fully phased out yet, and this change seems innocuous, so I'll just merge it in. Thanks!

PR #338, opened 2021-07-07
Add image labels to metadata

martinxsliu:
This PR adds the image's labels as a comma separated string to the version's metadata. Labels contain valuable metadata and in our case we encode information about the git commit that produced the image. This is useful information to display in the UI when viewing the build logs as well as the version list page when users they can use the labels to enable/disable/pin versions.

I have tested that the `$labels` variable contains the correct comma separated string when there are labels (see screenshot) and is an empty string when there are not.

image

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -76,6 +76,7 @@ if [ "$skip_download" = "false" ]; then
 
   echo "$image_id" > ${destination}/image-id
   docker inspect $image_id > ${destination}/docker_inspect.json
+  labels="$(cat ${destination}/docker_inspect.json | jq -r '.[0].Config.Labels // {} | to_entries | map("\(.key)=\(.value)") | join(", ")')"
 
   docker run \
     --cidfile=/tmp/container.cid \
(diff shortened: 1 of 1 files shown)


assets/in
```diff
@@ -76,6 +76,7 @@ if [ "$skip_download" = "false" ]; then
 
   echo "$image_id" > ${destination}/image-id
   docker inspect $image_id > ${destination}/docker_inspect.json
+  labels="$(cat ${destination}/docker_inspect.json | jq -r '.[0].Config.Labels // {} | to_entries | map("\(.key)=\(.value)") | join(", ")')"
```
xtremerui, line 79: what happens if the jq parsing encounters problem ? would it be always make labels empty only?

Could you add a test to make sure all cases are considered so it wont errored out?

xtremerui (commented):
Hi thx for the PR. This resource is depracated in favor of registry-image-resource. Ideally we don't want to pull in new feature here. Have you tried using registry-image-resource to archieve the same thing?

taylorsilva:
Closing since author never followed up on Rui's comment

PR #343, opened 2022-03-10
fixing docker login issue due 2 logins we are performing during put

niharbabu99:
When we are building and pushing a docker image via the resource we faced login error issue. Up on investigation it is something related to this [ISSUE](https://github.com/docker/hub-feedback/issues/1250). As per the issue, it says login fails for 2 reasons. 1. Due to password having special chars and 2. login is happening V2 [this gets solved via logout and login again].

As per the analysis I have done, looks like first [login](https://github.com/concourse/docker-image-resource/blob/master/assets/out#L55) is happening/happened via V1 and during second [login](https://github.com/concourse/docker-image-resource/blob/master/assets/out#L250) it tried via V2 and failure is seen.

image

image

image

image

sometime I am seeing the issue while performing get after successful PUT or at first login of the PUT. 
Though the containers used for PUT and GET are different, somehow there is login that happened already in these containers and it ends up in failing as it tries to login as part in/out script.

Failure for GET image post PUT
image

Failure for First Login for PUT
image

I think doing a logout before logging in is a safe option to avoid this situation. 

Request you to review and merge this PR.

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -117,7 +117,7 @@ log_in() {
   local username="$1"
   local password="$2"
   local registry="$3"
-
+  docker logout
   if [ -n "${username}" ] && [ -n "${password}" ]; then
     echo "${password}" | docker login -u "${username}" --password-stdin ${registry}
   else


niharbabu99:
@xtremerui can you please review this and get it merged. Its blocking us currently.

xtremerui:
Originally I think if we can tell whether docker has already logged on a registry then we could avoid doing logout/login again. But it seems like there is no an easy and clean way for that cross docker version, platform and OS.

Since we have creds available already, doing a `docker logout` seems reasonable.

@niharbabu99 have you tried building a docker-image-resource image with this PR first to solve your issue?

niharbabu99:
> Originally I think if we can tell whether docker has already logged on a registry then we could avoid doing logout/login again. But it seems like there is no an easy and clean way for that cross docker version, platform and OS.
> 
> Since we have creds available already, doing a `docker logout` seems reasonable.
> 
> @niharbabu99 have you tried building a docker-image-resource image with this PR first to solve your issue?

I haven't done that. If you want me to build a docker-image-resource and test it, I can do it and come back,

niharbabu99:
> @niharbabu99 please do if possible. I haven't noticed this error in our own CI so I would like to verify it will fix your issue.

I have validated my changes in our CI pipeline and I am not seeing any login error post my fix. I have run it multiple times and all the build starting from 60 are run with the docker-image I have created.
image

xtremerui (approved):

xtremerui:
Thank you for the PR.

PR #345, opened 2022-06-17
add support for authenticating to additional docker registries

mstandley-tempus:
This PR adds a new optional config parameter `additional_private_registries`, which allows a pipeline to authenticate to multiple docker registries.  

This is useful when trying to build a Dockerfile that depends on an image that is only available from a private registry, which is separate from the registry where the resulting docker image will be published.  For example, pulling from a private 3rd party docker registry and publishing to an internal docker registry, or transitioning to a new docker registry while still allowing pulls from an old docker registry.  

I have built this resource locally and tested it with our Concourse, per the instructions in the README.  When specifying additional registries, the resource successfully authenticated to all of them, and published the image as expected.  The resource still works as expected when no additional registries are configured.  

My only concern is with the workaround to the dockerhub login problem, introduced in https://github.com/concourse/docker-image-resource/pull/343.  This change always runs `docker logout` prior to calling `docker login`, which prevents `additional_private_registries` from working.  So, I read the PR and the linked issue carefully, and I think I've come up with a more surgical way of using `docker logout` to work around the problem without removing the credentials set by `additional_private_registries`.  However, I can't reproduce that issue in the first place, so I can't be sure that this workaround is still effective for people who face this problem with dockerhub.  

Other than that, I think this change is ready to go!  Thanks for reading my pull-request!

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -118,12 +118,6 @@ log_in() {
   local password="$2"
   local registry="$3"
 
-  # workaround for docker hub login problem https://github.com/concourse/docker-image-resource/pull/343 https://github.com/docker/hub-feedback/issues/1250
-  if [[ -z ${registry} ]]; then
-    # if $registry is empty, that implies that we are trying to auth to docker hub, and we need this workaround
-    docker logout https://index.docker.io/v1/
-  fi
-
   if [ -n "${username}" ] && [ -n "${password}" ]; then
     echo "${password}" | docker login -u "${username}" --password-stdin ${registry}
   else
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -117,7 +117,13 @@ log_in() {
   local username="$1"
   local password="$2"
   local registry="$3"
-  docker logout
+
+  # workaround for docker hub login problem https://github.com/concourse/docker-image-resource/pull/343 https://github.com/docker/hub-feedback/issues/1250
+  if [[ -z ${registry} ]]; then
+    # if $registry is empty, that implies that we are trying to auth to docker hub, and we need this workaround
+    docker logout https://index.docker.io/v1/
+  fi
+
   if [ -n "${username}" ] && [ -n "${password}" ]; then
     echo "${password}" | docker login -u "${username}" --password-stdin ${registry}
   else
--- a/README.md
+++ b/README.md
@@ -24,6 +24,23 @@ Note: docker registry must be [v2](https://docs.docker.com/registry/spec/api/).
 
 * `password`: *Optional.* The password to use when authenticating.
 
+* `additional_private_registries`: *Optional.* An array of objects with the
+  following format:
+
+  ```yaml
+  additional_private_registries:
+  - registry: example.com/my-private-docker-registry
+    username: my-username
+    password: ((my-secret:my-secret))
+  - registry: example.com/another-private-docker-registry
+    username: another-username
+    password: ((another-secret:another-secret))
+  ```
+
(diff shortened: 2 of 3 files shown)


mstandley-tempus:
> > Greetings @niharbabu99! Will this change to your `docker logout` workaround be OK?
> 
> Looks good to me.

Thank you!

assets/common.sh
```diff
@@ -117,7 +117,13 @@ log_in() {
   local username="$1"
   local password="$2"
   local registry="$3"
-  docker logout
+
+  # workaround for docker hub login problem https://github.com/concourse/docker-image-resource/pull/343 https://github.com/docker/hub-feedback/issues/1250
+  if [[ -z ${registry} ]]; then
+    # if $registry is empty, that implies that we are trying to auth to docker hub, and we need this workaround
+    docker logout https://index.docker.io/v1/
```
xtremerui, line 124: Do you have a reference of the docker hub URL for log out?  How about v2?

mstandley-tempus: Alternatively, we can consider removing the 2nd `docker login` call here: https://github.com/concourse/docker-image-resource/blob/688f57658d1b187739ea407b784751a055794cc0/assets/out#L250

I think this was added in this PR https://github.com/concourse/docker-image-resource/pull/332/files

If I understand this right, we have to remove `credsStore` from `~/.docker/config.json` before running `docker build` to get the build to proceed if ECR credentials are specified.  When the build is over, we want to restore the credentials to `~/.docker/config.json`, so we run `docker login` again to do that.  It is during this 2nd docker login that we get the bad behavior for which the https://github.com/concourse/docker-image-resource/pull/343 workaround is needed.

But, wouldn't it be more expedient to run `cp ~/.docker/config.json ~/.docker/config.json.bak` before `docker build` and `mv ~/.docker/config.json.bak ~/.docker/config.json` to restore the credentials instead of running `docker login` again?

mstandley-tempus: Greetings @ericb-summit! I have a question about your PR https://github.com/concourse/docker-image-resource/pull/332.

I want to avoid calling `docker login` a 2nd time.  Do you think it would work to backup `~/.docker/config.json` prior to removing `credsStore` then restore it after `docker build` finishes?

ericb-summit: It should work.  I did my PR as I did to minimize the changes to the code base.  The `config.json` after logout + login should be exactly as it was before, because the way `docker login` stores creds is mind-numbingly dumb; just `base64(user:pass)` in the relevant `auths` section of `config.json`.

mstandley-tempus: > It should work. I did my PR as I did to minimize the changes to the code base. The `config.json` after logout + login should be exactly as it was before, because the way `docker login` stores creds is mind-numbingly dumb; just `base64(user:pass)` in the relevant `auths` section of `config.json`.

Thank you for confirming that for me!

mstandley-tempus: I'm marking this conversation as resolved, because the line that needed a reference has been removed.  The docker logout workaround is no longer needed, because the 2nd docker login has been removed.

mstandley-tempus:
I've repeated my tests after the latest change in b45b9daae8a15b59caaa9ed00d53ca03994e9993 removing `docker logout` and removing the 2nd `docker login` that precipitated the need for it.  Everything still seems to be working as expected!

xtremerui (approved):
LGTM! Thx!

pc-jedi:
@xtremerui @niharbabu99  Could you create a new release containing that feature?

PR #352, opened 2023-02-17
fix additional_private_registries on alpine

mstandley-tempus:
The `additional_private_registries` parameter (from this PR: https://github.com/concourse/docker-image-resource/pull/345)  is broken when when this resource type is built using the alpine linux Dockerfile, but it works as intended using the ubuntu Dockerfile.  

The reason for this is because alpine's `base64` doesn't accept `--decode`.  Both ubuntu and alpine understand `-d`, so let's change it to pass that instead.  

Crucially, the docker-image resource type bundled with concourse is built from the alpine Dockerfile, so `additional_private_registries` will not work until this is merged, released, and bundled into a new version of concourse.

### DIFF
--- a/assets/out
+++ b/assets/out
@@ -59,9 +59,9 @@ additional_private_registries_base64=$(jq -r '.source.additional_private_registr
 
 # authenticate to additional registries (if any)
 for base64_line in ${additional_private_registries_base64}; do
-  additional_registry=$(echo $base64_line | base64 --decode | jq -r '.registry')
-  additional_username=$(echo $base64_line | base64 --decode | jq -r '.username')
-  additional_password=$(echo $base64_line | base64 --decode | jq -r '.password')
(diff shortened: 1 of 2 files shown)


mstandley-tempus:
@xtremerui can you please review my PR?

xtremerui:
Hi thank you for the PR. Is it possbile to add a test here https://github.com/concourse/docker-image-resource/blob/fff93fbac057429f8a0d80d6dba583ed199e8405/tests/out_test.go#L273?

mstandley-tempus:
> Hi thank you for the PR. Is it possbile to add a test here?
> 
> https://github.com/concourse/docker-image-resource/blob/fff93fbac057429f8a0d80d6dba583ed199e8405/tests/out_test.go#L273

yes @xtremerui -- added!

xtremerui (approved):
LGTM!

PR #360, opened 2023-08-24
fix: missing Docker buildx plugin

### LINKED ISSUE #359: docker_buildkit doesn't work?
### Describe the bug

I've enabled `docker_buildkit: 1` in the `put` step of a `docker-image` resource in my pipeline config and getting the following error:

```
ERROR: BuildKit is enabled but the buildx component is missing or broken.
       Install the buildx component to build images with BuildKit:
       https://docs.docker.com/go/buildx/
```

If I `fly hijack` the container and run `apt-get update` and `apt-get install docker-buildx-plugin` this makes the error go away, which suggests to me the plugin is missing from the docker image.

Looking at the [Dockerfile in this repo I see that it installs docker but not the plugin](https://github.com/concourse/docker-image-resource/blob/master/Dockerfile#L45), so it might be as simple as adding it to the list of things to install? I'm under the assumption this is the image the `put` step runs in, but it's highly possible I have a flawed mental model of what is happening though as I'm still getting my head around how all the various pieces fit together.

### Reproduction steps

1. Add `docker_buildkit: 1` to the `params` in the `put` step of a docker-image resource
2. Set the pipeline changes
3. Try and run the pipeline

### Expected behavior

The put step works with `docker_buildkit: 1` enabled

### Additional context

_No response_

TheDevMinerTV:
This fixes #359 by just installing the Docker Buildx plugin.


xtremerui (approved):
Thank you for the fix. Not sure why the original PR that introduced the param in put step didn't include this. It might be that buildx was later extracted out from docker itself so now it needs to be exclusively installed?

PR #363, opened 2024-04-05
Add secrets support

JasonDictos:
Add support for using the secret api in docker through the concourse ci yaml


JasonDictos:
Replaced by: https://github.com/concourse/docker-image-resource/pull/366

PR #368, opened 2024-07-24
Support GCR mirror that send application/vnd.oci.image.index.v1+json

mariash:
GCR mirror requires [application/vnd.oci.image.index.v1+json](https://github.com/opencontainers/image-spec/blob/main/image-index.md) media type in Accept headers. Otherwise, you will get an error in docker-image-resource that client doesn't accept that media-type and the check command won't be able to fetch the manifest digest.

We were having an issue running the test since the supported ginkgo is outdated, please verify and fix as needed.


xtremerui (approved):
LGTM!

xtremerui:
merging as the test already passed in previous run.

